PR #44, opened 2019-01-30
Fixes #27, fixed top with css

### LINKED ISSUE #27: Keep the top information while scrolling
This is more of an enhancement rather than an issue.
When browsing a folder, if I want to go back, I have to go to the top of the page in order to be able to navigate in the folder structure.
It would be nice to have it always available on the top of the page (top menu, navigation on the left and map on the right). Only the fotos would be scrolled.

(BTW, excellent new feature with the latest release. Thanks for it!!)

softbrix:
A simple solution on an always visible navigation menu.


bpatrik:
Hi,

sorry for the delay. I checked your solution. It looks good, I like it.
Unfortunately it feels to be a half solution for me (since bar takes up double space, especially on phones that it would require).
The app requires some refactoring to merge the two bars. I also would like to make the bar, disappear when you scroll down on a phone and only show when you scroll up a bit.
I'll merge your solution once I'll find the weekend to do this refactoring so that you will show up at the contributors. 
Sorry for not adding it as it is, but I would like the app to reflect my best effort on developing such an application.

softbrix:
Yes, this was the quick and somewhat dirty fix. 
My suggestion for the phone ui is to place the navigation in the bottom instead, like the navigation in the Instagram app. Then your users can easily access the navigation with one hand. You will need some more css and javascript logic based on what kind of device and the width of the screen though. 
And a couple of nice looking icons instead of text, since you don't have the same space on a phone.
I don't think hiding the navigation while scrolling in the gallery is necessary as the user are navigating and want easy access to filters and breadcrumbs. On the other hand while opening a single image then the user has shifted focus and is only interested in what they are focused on. But I guess that this already works today.

bpatrik:
Latest commit should support this.

PR #63, opened 2019-04-11
Remove .toString() from disk error handling

bobobo1618:
For the most part, error messages look like:
```js
{ error: 
      { errno: -2,
        code: 'ENOENT',
        syscall: 'stat',
        path: '/data/path/stuff/place' } }
```

But `toString()` just turns it into `[object Object]`, which is quite useless as a log message.


bpatrik:
Thank you!

PR #102, opened 2019-09-22
Translation ro

Alex-Badea:
Added translation in Romanian language


bpatrik:
Sorry I did not see this before

Issue #87, opened 2019-07-17
Suggestion: Hide irrelevant subfolders


occino:
Hi,
I'm very interessed in an option to hide irrelevant subfolders withing the images folder which does not contain any photos and videos. This way folders like '.Spotlight-V100' would not appear as well and leads to a better navigation.

Best regards from Berlin,
Nico

bpatrik:
Hi Nico, 

Sorry, but I'm not considering adding a feature like that, but feel free to jump in.
I'm reading the files and folders here: 
https://github.com/bpatrik/pigallery2/blob/f069aea742bb018d96a433bde2a04f87a5242b90/backend/model/threading/DiskMangerWorker.ts#L71
You can customize the app for your own taste.
If you feel to implement a versatile solution (like a ignore list in the config, or supporting .ignore file in the folders) I would consider merging it.

Best,
Patrik

bpatrik:
Hi,

Thanks to tmorley2000, this will part of the next release.

Best,
Patrik

nikolajhendel:
Hi, I'm trying to use this functionality to hide various folders (.recycle, .thumbnails, etc.) but can't seem to get it to work.
I assume the functionality is enabled by using the "Exclude Folder List" and "Exclude File List" under Index section in advanced settings?

I've tried to enable this in v 1.8 docker image (can only do via direct config.json edit) and 1.8.1 docker image nightly stretch (via config interface), but neither seems to work, as folders are still shown. I have also tried to re-run index after setting exclude rules (tried both folder name and file name) but doesn't hide folders in gallery view.

Am I misunderstanding something?

Cheers
Nik
p.s. thanks for a great gallery application!

tuxthepenguin84:
Hi @nikolajhendel,

I ran into a similar issue as you, however what you need to do is run the indexing after adding a folder to the exclude list, it will then not show in the gallery. Hope this helps.

-Sam

bpatrik:
Yes, reindexing is needed after changing the exclusion list.

hajo62:
Did a reindexing but I still see _all_ folders, although I have added a .ignore file to most root folders (1970-1979, ...) and after that did not work added those to the Exclude Folder List:
<img width="1281" alt="grafik" src="https://user-images.githubusercontent.com/15800980/147921888-2c76b5ec-2dde-4a67-886f-e518a601b1fe.png">

hajo62:
I just triggered the re-indexing but it still does not work.
[...]

PR #109, opened 2019-12-04
Ability to exclude directories, and config to choose them.

tmorley2000:
This provides a mechanism to exclude folders from the index. Both by name, and by folders that contain a particular named file.

This should be a fix for issue #87


bpatrik:
Looks good!

Thank you!

Patrik

PR #110, opened 2019-12-12
Update alpine dockerfile

Suika:
Since I don't know how to generate the config file without launching the server, this should do. Since it's only a build process, shouldn't matter too much.

Also, please add the alpine build to the docker hub autobuild process.

### DIFF
--- a/docker/alpine/Dockerfile
+++ b/docker/alpine/Dockerfile
@@ -8,7 +8,7 @@ RUN mkdir -p /build/release/data/config && \
     mkdir -p /build/release/data/db && \
     mkdir -p /build/release/data/image && \
     mkdir -p /build/release/data/TEMP && \
-    cd /build/release && node backend/server.js && \
+    cd /build/release && node src/backend/server.js || true && \
     sed -i 's/demo/data/g' config.json && sed -i 's@sqlite\.db@data/db/sqlite\.db@' config.json && \
     mv /build/release/config.json /build/release/data/config/config.json
 


bpatrik:
Hi,

Sorry I was a but busy in the last 6-8 months, so I could not put much effort in the project. Therefore I just merged your docker previously. It seems that you have a better understanding on Docker than I do, so now I would like to have some question about it (here inline, as I cannot review past commits): 

1) 
https://github.com/bpatrik/pigallery2/blob/2ac0eaa57a82c8a7374d213188b2df126a53a138/docker/alpine/Dockerfile#L5-L6
Here you build the project, create the release, then install the release. Now you have the chance to remove the 'build' folder. (It contains all dev dependencies that is required for running the tests and building angular.) Or this line `COPY --from=0 /build/release /app` solves it and it wont be present in the running container?


2)  ` cd /build/release && node src/backend/server.js || true && \`
I recommend using `npm start` instead. That should always point to the correct script. Also what does `|| true` stand for?
With the latest [commit](https://github.com/bpatrik/pigallery2/commit/2757af6bf0f9789284aaa2c8d295db3f6b320bd6) I made a `--config-only` switch that creates the config file and exists the app. See [readme](https://github.com/bpatrik/pigallery2#13-advanced-configuration).

3) What are the benefits that you have separate building and running docker? 

+1) Do you happen to have experiences building arm based image at Docker hub? 
I only found this: https://collabnix.com/building-arm-based-docker-images-on-docker-desktop-made-possible-using-buildx/ ,but it requires manual build and push to the hub.

Thank you!

Best,
Patrik

Suika:
This is basicall a staged build, every build stage starts with "FROM" and the last stage is the stage that is tagged and uploaded. They can be named, but with a small build it does not matter https://docs.docker.com/develop/develop-images/multistage-build/.

1. And to answer your question, yes the `COPY --from=0 /build/release /app` copies the relase from the container it was built on to the new container.

2. `command1 || command2` if command1 fails, command2 will execute, in this case a command true, which prevents the pipe from breaking. Will change it to `--config-only` shortly.

3. Every layer in docker is RO except the last one that is created when you start the container. Meaning if you you are to do:
```
RUN npm install --unsafe-perm
RUN npm run create-release
RUN cd /build/release && npm install --unsafe-perm
RUN cp /build/release /app && rm -rf /build
```

even though the end result that you will see when starting the container on your system is that you don't see the folder, that folder with all it's data exists in the intermediate layers. Meaning you would download more data than you actually need. Or say you need some dev packages to compile something but don't need that stuff in the relase contianer (python)

Yes you can do `RUN npm install --unsafe-perm && npm run create-release && cd /build && release && npm install --unsafe-perm && cp /build/release /app && rm -rf /build`

But then comes the caching, do you want to use buld caching or not. Depending on how you write the dockerfile, it will speed up a few things drastically. In this case, were I to change something in the lower part of the dockerfile(realease stage), the whole thing would not need to compile the whole application, only to change... say the healthcheck. But if code were to change, the COPY would notice a change and drop the caching, forcing the first container to build the application.

And then there is also the whole update procedure for the container. If done properly you will only need to pull a few layers for an update, instead of all the layers. blablablabla

There are more reasons, but it comes down to: A clean container, without extra packages, without layers that are not needed, only has what it needs and does only that one thing.


> Do you happen to have experiences building arm based image at Docker hub?
you can't, docker hub does not support this arch
but you should be able to do it with github actions, example: https://github.com/Difegue/LANraragi/blob/dev/.github/workflows/push-continous-delivery.yml https://github.com/Difegue/LANraragi/tree/dev/tools/build/docker-multiarch

bpatrik:
Thank you for the clarification and the link.
I have added `--config-only` to the Dockerfile and added alpine to dicker hub.

PR #113, opened 2019-12-23
Speedup docker builds

Suika:
GHA allows 20 concurrent jobs. So make use of it.
https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#usage-limits

### DIFF
--- a/.github/workflows/blank.yml
+++ b/.github/workflows/blank.yml
@@ -27,7 +27,46 @@ jobs:
         with:
           name: pigallery2-release
           path: release
-  build:
+  build-alpine:
+    runs-on: [ubuntu-latest]
+    needs: [create-release]
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v1
+      -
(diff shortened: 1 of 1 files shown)


bpatrik:
Hi Thanks
I was also thinking about that `release building` can be cached if no source files are changed
And ffmpeg install could be aslo cached in docker file (which seems that currently are not cached):
https://github.com/bpatrik/pigallery2/blob/3f7751024eb3156147fe264e404ad92d2a5c222d/docker/debian-stretch/Dockerfile.build#L22

Suika:
Since instances are created and destroyed, there is no cahce to be used. 
`--cache-from` https://github.com/docker/buildx#--cache-fromnametypetypekeyvalue should help.

Suika:
If it doesn't work as intended, you can always create your own base contianer and use that for the relase stage.

bpatrik:
thanks!

PR #130, opened 2020-01-27
Use ENV to set important paths

### LINKED ISSUE #124: Use alternate config.json file
If using different config.json files, you could configure several/different galleries easily.

npm start -- -conf alternate.conf

**Describe alternatives you've considered** (optional)
Couldn't find anything related in the documentation, maybe it has already been implemented.

Thxs!

Suika:
Makes it easier to change the options via ENV vars. Can be used to set the DB, IMAGE, TMP and CONF path.

### DIFF
--- a/docker/alpine/Dockerfile.build
+++ b/docker/alpine/Dockerfile.build
@@ -16,16 +16,20 @@ RUN mkdir -p /app/data/config && \
 #-----------------------------------------
 FROM node:12-alpine3.11 as main
 WORKDIR /app
+ENV NODE_ENV=production \
+    CONFIG_FILE=/app/data/config/config.json \
+    DB_PATH=/app/data/db \
+    MEDIA_PATH=/app/data/images \
+    TEMP_PATH=/app/data/tmp
 # command line arg orverride the config.json with these settings
 ENTRYPOINT ["node", "./src/backend/index",  \
   # after a extensive job (like video converting), pigallery calls gc, to clean up everthing as fast as possible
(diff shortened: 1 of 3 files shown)


bpatrik:
You can also directly set all configuration options through env variables (same way as cli, but without the `--` prefix. Case sensitive) like: 
```
ENV Server-Media-tempFolder=/app/data/tmp
````
(does not work with `--config-path`)

Also why would some want change the location of these files within the container? (this is not the first time, but I can't see the reason as one should bind with volumes)

Suika:
Depends. Having a single folder for multiple paths, same with databases.
`/app/data/db/a/`, `/app/data/db/b/`, `/app/data/db/c/`. Same with configs.

In case of #124, he probably wanted to use a single install to launch multiple instances. Which is valid.

> but I can't see the reason as one should bind with volumes

Depends on the persons setup. I also don't like if in certain cases a container creator tells me to use a his thought out layout when I can transparently mount my paths to the docker and have a container display the same paths I have on my host. Not the case with PiG2, but if there is an option and the users want to use it, let them.

If say that guy launched instaces via systemd, it would be easier to configure things through multiple configs and env variables that can be loaded dynamically, etc.
Docker is prefered setup and the supported setup. Some derivations can be accounted to being problems.

Meh, your choice.

bpatrik:
Okay. I see. I'm not against it, just did not got it earlier.
Also I'm working on a mayor config update that will work better with docker.

PR #140, opened 2020-03-03
Dark theme resolves #68

### LINKED ISSUE #68: Suggestion: Offer a dark theme
Thanks a lot for your effort on pigallery2. I use it on a AMD64 NAS with Debian 9.

I would like to suggest a to implement a dark theme additional to the the white (light) theme. So a user could choose between the two themes.

barry-luijten:
First implementation of a dark theme for PiGallery. 
The theme is based on switching values of CSS variables.
Currently, two themes are hardcoded into theme.service.ts: themeLight (the default) and themeDark.
The following CSS variables are (re)defined and initialized in the new Theme class:
```
        this.vars.push(new ThemeVariable('--text-color'));
        this.vars.push(new ThemeVariable('--background-color'));
        this.vars.push(new ThemeVariable('--background-color-secondary'));
        this.vars.push(new ThemeVariable('--text-color-secondary'));
        this.vars.push(new ThemeVariable('--text-color-disabled'));
        this.vars.push(new ThemeVariable('--divider-color'));
        this.vars.push(new ThemeVariable('--blue','#94c5f9'));
        this.vars.push(new ThemeVariable('--primary','var(--blue)'));
```
Both themes are instances of the Theme class and can be applied using the exported methods of the ThemeService class:
- setThemeLight()
- setThemeDark()

The active theme is configurable via the Settings menu and stored in config.json. There is a new section in Settings named "User Interface", which now contains one boolean setting: darkMode.
We may want to extend this later on to support multiple themes. We could also add the option to normal users to change the theme at runtime because CSS variables are dynamic and don't require any compilation.

Most of the html and css files have been touched to use the new CSS variables. 

I took the liberty of tweaking the main navigation bar a little, so it aligns nicer, especially when it's collapsed.

Please let me know what you think!


bpatrik:
Hi,

sorry for the super late reply
Thank you very much for your PR.

I had the chance to check it. I noticed that you also changed the original color setup too.
I like that change at some places (empty directory icons probably look better now), but find it disturbing at some other (photos info bar is harder to read now). Also under the duplicate tab, the original theme stayed intact. So the design is not the same now across the whole page. 

Here are some screenshots:

infobar
main

Would you have the capacity to make the original theme as it was before and only change the dark theme? If not that's fine too. I can do that too at some point

Issue #163, opened 2020-06-23
mysql - fallback to memory db allows admin:admin login!


soneji:
So I want to set up with mysql and docker compose, but the app server starts up too quick for the database server, causing it to revert to memory db, where username and password admin:admin work. This is a major security vulnerability. 
```
Error during initializing SQL falling back temporally to memory DB
"Error: ER_ACCESS_DENIED_ERROR: Access denied for user 'pigallery2'@'192.168.112.3' (using password: YES)"
```

If I log in with admin:admin and resubmit the database information, this is a temporary fix, but every time I restart the server, this issue persists. This error is inherant to the nature of MYSQL's slower start up time. 

A solution would be to not start the app until a connection can be made to the database, once the MYSQL database has been chosen. Additionally displaying an error message on the home screen like "cannot log in, because database has not yet been connected". 

Using docker-compose on Linux. FInd below my docker-compose.yml:
```yaml
version: '3'
services:

  db:
    image: mariadb
    volumes:
      - ./db-data:/var/lib/mysql
    environment:
      - MYSQL_ROOT_PASSWORD=_________________
      - MYSQL_USER=pigallery2
      - MYSQL_PASSWORD=pigallery2_pass
      - MYSQL_DATABASE=pigallery2

  app:
    image: bpatrik/pigallery2:latest
    command: --Server-Database-mysql-host=db --Server-Database-mysql-username=pigallery2 --Server-Database-mysql-password=pigallery2_pass--Server-Database-mysql-database=pigallery2
    environment:
      - NODE_ENV=production
    volumes:
      - "./config:/app/data/config"
      - "./db-data:/app/data/db"
      - "/my/images:/app/data/images/"
      - "/tmp/pigallery/:/app/data/tmp" 
    ports:
      - 8008:80
    restart: always
    depends_on:
      - db
```
Find below my docker-compose logs:
```bash
app_1  | using config path:/app/data/config/config.json
app_1  | 6/23/2020, 6:30:38 PM[INFO] [server] running diagnostics...
app_1  | 6/23/2020, 6:30:38 PM[WARN] [ConfigDiagnostics] [SQL error] Error: ER_ACCESS_DENIED_ERROR: Access denied for user 'pigallery2'@'192.168.112.3' (using password: YES)
app_1  | 6/23/2020, 6:30:38 PM[WARN] [ConfigDiagnostics] Error during initializing SQL falling back temporally to memory DB
[...]

PR #164, opened 2020-06-25
make app wait for db to start up first

overclockedllama:
Fix for #163 
Prevents using memory database if db is still starting up

The server still starts up on first start with a sqlite db, but once you configure it to use mysql from the GUI, it uses that.

### DIFF
--- a/docker/docker-compose/with-mysql/docker-compose.yml
+++ b/docker/docker-compose/with-mysql/docker-compose.yml
@@ -26,7 +26,7 @@ services:
 
   pigallery2:
     image: bpatrik/pigallery2:latest
-    command: --Server-Database-mysql-host=pigallery-db --Server-Database-mysql-username=pigallery2 --Server-Database-mysql-password=pigallery2_pass--Server-Database-mysql-database=pigallery2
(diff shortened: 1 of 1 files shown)


bpatrik:
I did not have the chance to try it.
Can you confirm that it works properly. I suspect that the following command will be executed:
`node ./src/backend/index --expose-gc  --config-path=$CONFIG_FILE  sh -c 'bin/wait-for pigallery-db:3306 -- --Server-Database-mysql-host=pigallery-db --Server-Database-mysql-username=pigallery2 --Server-Database-mysql-password=pigallery2_pass--Server-Database-mysql-database=pigallery2'`
And I think that would fail, as the dockerfile has an entypoint:
https://github.com/bpatrik/pigallery2/blob/64b85930353319d30b757aca9b6d19c219fe1cd7/docker/alpine/Dockerfile.build#L29
to what the `command:` gets concatenated.

overclockedllama:
I have tested it and it indeed works. That is not the command being run. 
The `--` is a unix shorthand for STDIN where the entrypoint `node ./src/backend/index --expose-gc --config-path=$CONFIG_FILE` is being passed in as STDIN, hence the command is: 

`sh -c 'bin/wait-for pigallery-db:3306 -- --Server-Data...` =>
`sh -c 'bin/wait-for pigallery-db:3306 STDIN --Server-Data...` =>
`sh -c 'bin/wait-for pigallery-db:3306 node ./src/backend/index --expose-gc --config-path=$CONFIG_FILE --Server-Data...`

bpatrik:
Thank you, perfect!
i was not aware the functionality of the `--` .

Issue #154, opened 2020-05-03
Recursive random photo is not recursive


BbillyBob:
**Describe the bug**
Files tree in the PiGallery (with /app/data/images/ as root index folder) :
/app/data/images/galleries/
/app/data/images/galleries/0.jpg
/app/data/images/galleries/dir1/
/app/data/images/galleries/dir1/1.jpg
/app/data/images/galleries/dir2/
/app/data/images/galleries/dir2/2.jpg

_Expected_
1. With http://mygallery.local/api/gallery/random/ : get alternatively 0.jpg, 1.jpg or 2.jpg when refreshed
2. With http://mygallery.local/api/gallery/random/?directory=galleries : get 0.jpg even when refreshed 10 times
3. With http://mygallery.local/api/gallery/random/?directory=galleries&recursive=true : get alternatively 0.jpg, 1.jpg or 2.jpg when refreshed

_Reality_
1. OK
2. OK
3. Got only 0.jpg
3.bis If I delete 0.jpg, I get the following error :
{"error":{"code":12,"message":"No photo found","detailsStr":""},"result":null}
I get the same error with 2. after 0.jpg deletion (but it is expected in this case).
So it seems the recursive parameter does not work or I missed something.

Additionally (you may want me to create another bug issue for this), the "&recursive=true" in the URL field is not updated when I switch "recursive" to on/off in the Random Link menu in the GUI. It is updated when I change another field (Date or Resolution) though.

**Environment**
 - Latest docker version without nginx proxy
 - Browser Firefox and Chrome

bpatrik:
Thanks for the report. I'll look into it.

bpatrik:
The last commit should solve this.

PR #165, opened 2020-06-29
Folder preview image will be taken from all child directories not only direct childs

lexas2509:


lexas2509:
Inspired by this issue [https://github.com/bpatrik/pigallery2/issues/154] this MR is fixing blank folder thumbnail while there are the non-empty subfolders (but the folder itself doesn't have any images)

bpatrik:
I looked at the code. 

I have to double check, but I think this approach leads to inconsistent UI.
Let assume you have this dir setup:
+DirA
   +-a.jpg
   +-DirB
        +--DirC
              +-c.jpg

The inconsistency can come from showing c.jpg in DirB (not only for preview).

bpatrik:
After I gave some thought to it. You might not notice the inconsistency, but I still find this solution a bit misleading (namely you use a variable (.media) to store a image in it that is not part of a given directory. 

A more proper solution would be to add a new filed to the directory here: https://github.com/bpatrik/pigallery2/blob/3deb8e26f5cda166d32c34188d12583ab00f6272/src/common/entities/DirectoryDTO.ts#L14
like
`previews: S[];`
And store the previews for a given directory there. Should not be too hard to implement, just a lot of refactoring.

src/backend/model/diagnostics/ConfigDiagnostics.ts
```diff
@@ -30,7 +30,7 @@ export class ConfigDiagnostics {
     if (databaseConfig.type !== ServerConfig.DatabaseType.memory) {
       await SQLConnection.tryConnection(databaseConfig);
     }
-    if (databaseConfig.type !== ServerConfig.DatabaseType.sqlite) {
+    if (databaseConfig.type === ServerConfig.DatabaseType.sqlite) {
```
bpatrik, line 33: This is a really nice catch. I would e happy to merge this change even separately from the other changes.

lexas2509: I can prepare a separate change for this if you want.

snoopyCode: I've tested the actual fix several times and it works fine. No inconsistencies found. Sure, it would be better to have something like a dedicated `previews` attribute, however it seems to be a lot of work.

I would also suggest to move the fix of 'ConfigDiagnostics.ts' into another pull request and merge the actual one. @lexas2509 Can you do it in your pull request, otherwise I can do another one to get the fix soon merged :-)

bpatrik: Alright. Fingers crossed. I'll merge this.
Worst case the nightly build wont work. There isn't much in it compared to 1.8.2. 
I wanted to fix this actually properly. Now I will be forced to. 
I might just remove the memory (non DB) approach. That will make things simpler and with docker, having at least an sqlite DB, should not be an issue. Although, I'm not sure how may people are using it.

snoopyCode: Thanks! Tested the result of the nightly build and it works quite well on my side.

PR #176, opened 2020-09-02
Fix dimesions of rotated videos in metadata

axlbdev:
This will apply rotation transformation from `ffprobe` to metadata's size. Without it vertical videos (e.g. from phone camera) appears stretched in gallery.
`video_rotate.mp4` asset created with command `ffmpeg -i video.mp4 -map_metadata 0 -metadata:s:v rotate="90" -codec copy video_rotate.mp4`

Before:
before

After:
after


bpatrik:
Awesome, thank you!

PR #182, opened 2020-09-13
Add web manifest

tuur29:
Will allow you to create better home screen and desktop shortcuts on desktop and mobile Chromium browsers.
See: https://web.dev/add-manifest/


bpatrik:
Thank you for your contribution!

PR #189, opened 2020-10-13
Use object-fit: cover in gallery grid

### LINKED ISSUE #184: Incorrect aspect ratio displayed for thumbnails
**Describe the bug**
I've been using the docker nightly build and have noticed recently that the incorrect aspect ratio is being used for thumbnails. I switched over from the `latest` docker build but noticed it had memory issues but aspect ratio was displayed correctly with that version.

Unfortunately, I don't have any particular reproduction steps, it just seems to happen with most of my pictures (although some are displayed correctly for some reason)

**Photo/video (optional) that causes the bug**
An example photo:

**Screenshots**
However, the thumbnail is displayed with equal height and width
image

- I'm currently running PiGallery on a Raspberry Pi 4, 64-bit enabled within docker
 - I've been able to repro on both Chrome and Firefox (unable to load on safari)

**Used app version**:
- docker-nightly (specifically alpine)

Happy to provide any more details/logs I can. As far as I can tell, no errors appeared in the logs during thumbnail creation or indexing

lbischof:
I think this fixes #184. I did not test this locally, only in the developer tools of the browser.

### DIFF
--- a/src/frontend/app/ui/gallery/grid/photo/photo.grid.gallery.component.css
+++ b/src/frontend/app/ui/gallery/grid/photo/photo.grid.gallery.component.css
@@ -8,6 +8,7 @@ img {
   -o-animation: fadein 2s; /* Opera < 12.1 */
   animation: fadein 2s;
   vertical-align: unset;
+  object-fit: cover;
 }
 
 @keyframes fadein {


bpatrik:
I shell test this, but I have the feeling this does not produces the correct result. (See details in the bug)

bpatrik:
I had a quick look and it seems in Firefox it crops the photo, while in Chrome it increases the distortion. :/

PR #190, opened 2020-10-14
Replace old Mapbox tiles with Static Tiles API

matteodelabre:
Old Mapbox classic styles and API have been deprecated then removed since June 2020. This PR changes the default Mapbox tiles to request the new API.

See <https://docs.mapbox.com/help/troubleshooting/migrate-legacy-static-tiles-api/>.


bpatrik:
Thank you!

PR #194, opened 2020-10-22
Fix docker instructions for running without docker-compose

zertrin:
Correct definition of config volume. Previously, it would create a `config.json` folder on the host upon first creation of the container, which prevents the app to create the `config.json` file.


bpatrik:
Nice, thank you!

PR #195, opened 2020-10-25
Update Dockerfiles

### LINKED ISSUE #193: Healthcheck not working because wget is not installed in the docker image
**Describe the bug**
Using the latest nightly docker image, I noticed that the health check was not working properly, simply because there is no `wget` in the image.

Proposed solution: add `wget` inside the docker image.

**Screenshots**

image

 - OS: Host:  Linux Debian buster, running pigallery2 in docker without using docker-compose (because I already have a reverse proxy to terminate SSL on the host).

Here is my docker run command for reference:

```bash
#!/bin/bash

TAG="nightly-stretch"
DATA_ROOT_PATH="/data/pigallery2"
IMAGES_PATH="/data/redacted"

docker run -d --name pigallery2 --restart always \
    -p 8314:80 \
   -e NODE_ENV=production \
   -v "${DATA_ROOT_PATH}/config":/app/data/config \
   -v "${DATA_ROOT_PATH}/db":/app/data/db \
   -v "${IMAGES_PATH}":/app/data/images:ro \
   -v "${DATA_ROOT_PATH}/temp":/app/data/tmp \
   bpatrik/pigallery2:${TAG}
```

**Used app version**:
- docker-nightly

zertrin:
Some updates to the Dockerfiles

- added `wget` to Debian images (it was removed from the node base image, see https://github.com/nodejs/docker-node/issues/1185). Fixes #193 
- resynchronized the standalone Dockerfile with the one used for builds
- apt install without recommended packages
- clean apt cache after install to reduce image size a bit

Note:
- the edits to the two `Dockerfile.build` are simply intended to minimize the differences between all files (e.g. when using `diff` or `vimdiff`)
- I have only tested the standalone Dockerfile on my side, please test a bit on your side once the automated builds have completed.
- I have added the option `--no-install-recommends` to reduce the number of packages installed. It might (or might not) be an issue with video transcoding. I have tested that video transcoding runs without errors, but on a tiny sample. Please consider checking that nothing broke for wider use cases I could not test.

### DIFF
--- a/docker/alpine/Dockerfile.build
+++ b/docker/alpine/Dockerfile.build
@@ -14,7 +14,7 @@ RUN mkdir -p /app/data/config && \
 
 #-----------------MAIN--------------------
 #-----------------------------------------
-FROM node:12-alpine3.11 as main
+FROM node:12-alpine3.11 AS main
 WORKDIR /app
 ENV NODE_ENV=production \
     # overrides only the default value of the settings (the actualy value can be overwritten through config.json)
@@ -24,7 +24,6 @@ ENV NODE_ENV=production \
     # flagging dockerized environemnt
     PI_DOCKER=true
 
-
 EXPOSE 80
 RUN apk add --update-cache --repository https://alpine.global.ssl.fastly.net/alpine/v3.11/community/ \
     vips ffmpeg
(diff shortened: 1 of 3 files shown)


bpatrik:
Thank you!

PR #199, opened 2020-10-26
Add webpack dependency

lbischof:
My build seemed to fail without this. I'm not sure if it's needed though.


bpatrik:
Hi,
I should look into it, but I think angular should contain webpack.

lbischof:
It could well be that I messed something up, because I was trying to fix the other error. After deleting `node_modules` and rerunning `npm install` it seems to work. Sorry for wasting your time.

PR #224, opened 2021-02-02
Using natural-orderby for sorting

### LINKED ISSUE #223: Numeric String Sort
Unsure if this should be a Feature request or Bug report:

Sorting is currently done by comparing strings using `string.toLowerCase()`:

```typescript
      case SortingMethods.ascName:
      case SortingMethods.ascDate:
        this.directories.sort((a: DirectoryDTO, b: DirectoryDTO) => {
          if (a.name.toLowerCase() < b.name.toLowerCase()) {
            return -1;
          }
          if (a.name.toLowerCase() > b.name.toLowerCase()) {
            return 1;
          }
          return 0;
        });
        break;
```
(from https://github.com/bpatrik/pigallery2/blob/master/src/frontend/app/ui/gallery/gallery.component.ts#L175).

The problem is when numbers are involved:

```txt
Screenshot 1.jpeg
Screenshot 2.jpeg
Screenshot 3.jpeg
...
Screenshot 10.jpeg
```

would be sorted to
```txt
Screenshot 1.jpeg
Screenshot 10.jpeg
Screenshot 2.jpeg
Screenshot 3.jpeg
...
```

This is obiously not desired and the standard libs don't offer much help in this regard. There are multiple packages and solutions out there so it's not hard to fix but requires some discussion about what solution to use (if any at all).

erri120:
Resolves #223


bpatrik:
Thank you!

bruce30262:
Hello @erri120 @bpatrik , 

I found this PR only fixes the order in the gallery view, but not in the photo view ( the directories in `/gallery` are now in natural-order, but photos in each directory are still using lexicographic order )  

To fix this I think the following lines should also be updated : https://github.com/bpatrik/pigallery2/blob/7f321a3ff5ca16c855504f4b9995674d24f4e548/src/frontend/app/ui/gallery/grid/grid.gallery.component.ts#L230-L263  

Not sure if I have missed other files though, but I think that's it.

bruce30262:
@bpatrik I've created a PR ( #230 ) please check.

PR #230, opened 2021-02-26
Fix gallery order issue

bruce30262:
This PR fixes the order issue in the gallery while sorting directories/media files:  

* Add code so directories can be sorted by date  
The code in `src/frontend/app/ui/gallery/gallery.component.ts` shows that `SortingMethods.ascDate` and `SortingMethods.ascName` ( and `desc` as well ) are using the same code to sort the directories ( both sort by name in natural-order ).  I've added some code so directories now can be sorted by its `lastModified` attribute.  

* Add code so media files in a directories can be sorted in natural-order  
PR #224 added the code to sort the directories in natural-order, however the media files in directories are still using lexicographic-order. I've added some code in `src/frontend/app/ui/gallery/grid/grid.gallery.component.ts` to support natural-order for media files as well.  

Please do review this PR, since I'm not sure if the modification suits everyone needs. Discussion/suggestions are welcomed.

### DIFF
--- a/src/frontend/app/ui/gallery/gallery.component.ts
+++ b/src/frontend/app/ui/gallery/gallery.component.ts
@@ -172,12 +172,16 @@ export class GalleryComponent implements OnInit, OnDestroy {
     }
     switch (this._galleryService.sorting.value) {
       case SortingMethods.ascName:
-      case SortingMethods.ascDate:
         this.directories.sort((a: DirectoryDTO, b: DirectoryDTO) => compare()(a.name, b.name));
         break;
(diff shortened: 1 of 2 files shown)


bpatrik:
It looks good to me. 
I'm fine if we also sort directories based on date. I think that make sense. 
If you tested it and works, I can merge it.

bruce30262:
@bpatrik I've added one more commit to fix a minor bug ( forgot to add `break` in switch-case :P )   
The code has been tested and I think it's OK to merge

bpatrik:
Thanks!

PR #232, opened 2021-03-04
Add text shadow

tibrisch:
I have a really hard time reading names of faces and captions.
Especially if the photo is bright and/or grayscale.
Here is a simple fix.


bpatrik:
I can merge it know for a 'quick' fix.. Not a designer, but heard that shadow/outline is not cool lately.  A bubble like solution would be probably better, like what facebook does with the names and the arrows. Google photos does the same just with inverse color (dark background, light text/arrow color)

bpatrik:
oops, forgot to hit the merge button

PR #248, opened 2021-04-02
Search autocomplete suggestion on click

tibrisch:
I hope it's ok to add this too. I think it's nice to also have search on click.


bpatrik:
thanks, that should have been there.

PR #249, opened 2021-04-03
Show face region when zoom and pan.

tibrisch:
I am not expecting you to accept this code, in the current state at least. After much brute force and trying different things, I came up with this. This actually does the job well enough for me, however it's not pretty. What can I/we do to refine and make it acceptable? It took me a while to figure out that there is a dependency on photoAspect for the x axis drag. Also I have no clue where the number 9.45 comes from, it's something I found after trying lots of values. Any idea?

### DIFF
--- a/src/frontend/app/ui/gallery/lightbox/controls/controls.lightbox.gallery.component.html
+++ b/src/frontend/app/ui/gallery/lightbox/controls/controls.lightbox.gallery.component.html
@@ -62,8 +62,8 @@
         [style.transform]="ImageTransform"
         [style.top.%]="(zoom*(face.box.top+(face.box.height/2))-(activePhoto.gridMedia.Photo.metadata.size.height/2)*(zoom-1)) / activePhoto.gridMedia.Photo.metadata.size.height*100"
         [style.left.%]="(zoom*(face.box.left+(face.box.width/2))-(activePhoto.gridMedia.Photo.metadata.size.width/2)*(zoom-1)) / activePhoto.gridMedia.Photo.metadata.size.width*100"
-        [style.height.%]="face.box.height / activePhoto.gridMedia.Photo.metadata.size.height*100"
-        [style.width.%]="face.box.width / activePhoto.gridMedia.Photo.metadata.size.width*100"
+        [style.height.%]="face.box.height / activePhoto.gridMedia.Photo.metadata.size.height*100*zoom"
+        [style.width.%]="face.box.width / activePhoto.gridMedia.Photo.metadata.size.width*100*zoom"
         *ngFor="let face of activePhoto.gridMedia.Photo.metadata.faces">
         <div class="face-box"></div>
         <span class="face-name">{{face.name}}</span>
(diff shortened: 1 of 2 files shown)


src/frontend/app/ui/gallery/lightbox/controls/controls.lightbox.gallery.component.html
```diff
@@ -55,14 +55,14 @@
...
-         *ngIf="facesEnabled && activePhoto && zoom == 1 && activePhoto.gridMedia.Photo.metadata.faces && activePhoto.gridMedia.Photo.metadata.faces.length > 0">
+         *ngIf="facesEnabled && activePhoto && activePhoto.gridMedia.Photo.metadata.faces && activePhoto.gridMedia.Photo.metadata.faces.length > 0">
       <a
         class="face"
         [routerLink]="['/search', getPersonSearchQuery(face.name)]"
-        [style.top.%]="face.box.top / activePhoto.gridMedia.Photo.metadata.size.height*100"
-        [style.left.%]="face.box.left / activePhoto.gridMedia.Photo.metadata.size.width*100"
-        [style.height.%]="face.box.height / activePhoto.gridMedia.Photo.metadata.size.height*100"
-        [style.width.%]="face.box.width / activePhoto.gridMedia.Photo.metadata.size.width*100"
+        [style.top.%]=" ((face.box.top +(face.box.top -(activePhoto.gridMedia.Photo.metadata.size.height/2))*(zoom-1)) / activePhoto.gridMedia.Photo.metadata.size.height*100) + (drag.y/9.45)"
```
bpatrik, line 62: I think this calculation should be simpler. 

You probably only need to offset the height/width with https://github.com/bpatrik/pigallery2/blob/24942b2ee1a2de01048c1aa053d7b68a28b73d37/src/frontend/app/ui/gallery/lightbox/media/media.lightbox.gallery.component.ts#L41-L45
and probably convert that to percent.

src/frontend/app/ui/gallery/lightbox/controls/controls.lightbox.gallery.component.ts
```diff
@@ -53,6 +53,7 @@ export class ControlsLightboxComponent implements OnDestroy, OnInit, OnChanges {
   private timerSub: Subscription;
   private prevDrag = {x: 0, y: 0};
   private prevZoom = 1;
+  public photoAspect = 1;
```
bpatrik, line 56: Not sure, but I think you can do the calculation without depending on the aspect ration

bpatrik: I started writing up a contribution doc: https://github.com/bpatrik/pigallery2/blob/master/CONTRIBUTING.md (not much in it yet) to help contributors getting started.

tibrisch: It's rewritten now. Works much better. Thank you for your advice!
I had to do scale(1/zoom) on the face-name, otherwise the text would also be scaled.
I am more happy with this solution. What do you think?

tibrisch:
It's rewritten now. Works much better. Thank you for your advice!
I had to do scale(1/zoom) on the face-name, otherwise the text would also be scaled.
I am more happy with this solution. What do you think?

tibrisch:
Rewritten for 3rd time. Now I figured out how to do this without using css scale(). So now text and box border is not impacted. Also found out that when on my smartphone, its nice to see  controls when doing pan. Added fix for that, or should I open separate bug for that one?

bpatrik:
Hi,

I took some time and looked into it and found some simpler solution: https://github.com/bpatrik/pigallery2/commit/4ced2c32261ced2435e7d85f024f4c369dc61a51 I will go with that. I wont merge this commit, sorry about it, but thank you for your effort.

tibrisch:
Very well done! Much better.
I'm very happy to have this enhancement in!

PR #261, opened 2021-04-17
Finalized italian translations + robots.txt

Magneticdud:
Hi, sorry for the late PR but suddenly I was buried in work.

Finished all the string translation to italian, then added another commit for the robots.txt that disallow indexing from search engines


bpatrik:
Thank you!

Issue #267, opened 2021-04-21
please support XMP tag "subject"


kagahd:
I discovered PiGallery2 only yesterday and I love it!
However, many of my photos store individual labels in the XMP section "subject". Unfortunately pigallery2 is unable to read (and search) such data.

I adapted PiGallery2 to fit my needs as follows:
At line 204 of file ./src/backend/model/threading/MetadataLoader.js I added the following code:

```
if(exif.subject && exif.subject.value){
   metadata.keywords = [];
   for(let i=0; i < exif.subject.value.length; i++){
      metadata.keywords.push(exif.subject.value[i].value);
   }
}
```
This works well for me but as you can see, my code **replaces** the "keywords" by my "subject" values. This is of course not a good solution for somebody who uses "keywords" (I don't use keywords). Hence my feature request to implement "subject" XMP metadata properly as a new property. 

Please excuse my ignorance to do it by myself. I think to implement this properly, at least the data model, DTO's and tests need to be adapted which I feel I'm not capable of. Thanks for your understanding and your effort!

bpatrik:
Hi
Can you elaborate more about this subject XMP section? Why do you use it instead of the iptc keywords? Do they have different purpose? Are you using a tool that puts these tags to subject instead of the normal keywords section?

If it is a valid use, I do not mind adding it to the lib. We can merge the current keywords list with this list of subjects. 
If it is something that is not standard, I would prefer not adding to the code.

kagahd:
Hi @bpatrik 
I'm using the Android app "F-Stop Gallery Pro" to tag my photos.

After a bit googling, what the "subject" field is for, I found out that _**"Subject" is the XMP place for keywords**._ ([source](https://www.carlseibert.com/xmp-iptciim-or-exif-which-is-preferred/)) 

The [specification](https://www.iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#keywords) of the XMP tag  `dc:subject` confirmt this statement:

Row header | Specification
-- | --
Name  | Keywords
Definition | Keywords to express the subject of the image. Keywords may be free text and don’t have to be taken from a controlled vocabulary. Codes from the controlled vocabulary IPTC Subject NewsCodes must go to the "Subject Code" field.
Help Text | Enter any number of keywords, terms or phrases used to express the subject matter in the image.
Implementation Note(s) | Single values of this field should not be restricted to single words but must allow for phrases as well.
Label | Keywords
[...]

PR #270, opened 2021-04-23
#267 support XMP tag 'subject'

kagahd:
PR for supporting xmp subject as keywords.


src/backend/model/threading/MetadataLoader.ts
```diff
@@ -216,7 +216,13 @@ export class MetadataLoader {
               if (exif.Rating) {
                 metadata.rating = (parseInt(exif.Rating.value, 10) as any);
               }
-
+              if(exif.subject && exif.subject.value){
+                for(let i=0; i < exif.subject.value.length; i++){
```
bpatrik, line 220: add something like this here ` metadata.keywords = iptcData.keywords || [];`

test/backend/assets/xmp/xmp_subject.json
```diff
@@ -0,0 +1,24 @@
...
+    "fStop": 2.4,
+    "focalLength": 4.32,
+    "make": "samsung",
+    "model": "SM-G975F"
+  },
+  "creationDate": 1614703656000,
+  "fileSize": 7909,
+  "orientation": 1,
+  "rating": 50,
+  "Rating": 3,
```
bpatrik, line 14: having two ratings fiels looks suspicious to me.

bpatrik:
I added some comments.
You can try running the test locally too. Your IDE should be able to help with that (Webstorm does help for me). If not, you can try running it from terminal with something like this: `mocha test/**/*.js --grep "^MetadataLoader"` from the project root. (mocha need to be installled globally)

bpatrik:
I will look into the errors in more details. It is is possible, that with this change the galllery will be able to read more metadata from the files, than we first expected (like reading keywords from png file which does not have exif data)
So the failing test might be features instead of bugs, but I need to look at them more closely

kagahd:
Yes, you're right. My modification reads more metadata from the files while the tests did not reflected them. That's the case for the `test_png.png` file which has 2 keywords in the XMP `subject` section. However, I had to add `"keywords":[ ]` to the `test_png.json` because the keywords were **not** retrieved by the `ExifReader`. Do you know why? 

I found out that `metadata.keywords` is only initialized if the photo has IPTC data. If it's not, my code could not push the keywords to the (not existing) array. So I had to initialize it if not yet done. That's the reason why the keywords of my photo were not present.

The tests for the photo with special characters was failing because I wrongly used the `.value` instead of `.description` to get the value from the `subject` fields. I found out that I had to use `.description` to get right encoding. However, I still don't understand why the photo has a keyword `squirrel` in both IPTC and XMP but the json file to test against doesn't have it. Nevertheless the test passes.

To make the test pass, I removed my rating from my test photo. I found out that XnView added to the XMP meta data a value from 0 to 100 as `MicrosoftPhoto.Rating` besides the original tag `xmp.Rating` which goes only from 0 to 5. ExifReader overwrites the second last read value by the last read value. In my case, `MicrosoftPhoto.Rating` overwrote `xmp.Rating` which made the test fail.

Last but not least I did not understand why `npm test` failed locally on test `39) SearchQueryParser`, I found out that it has nothing to do with my code changes because this test fails also locally without my code changes. However, the test passes online.

Thanks @bpatrik to point me to `mocha` to to run the test locally. I didn't know about it and used locally `npm test` which takes probably much more time to execute than just to test the `MetadataLoader`. I'll try it out.

kagahd:
Btw. `mocha test/**/*.js --grep "^MetadataLoader"` just showed `0 passing (2ms)`, so it seems that it does not run any test.

`npm test -- --grep "^MetadataLoader"` did the trick but was as slow as `npm test` (1:23 min).

bpatrik:
Thank you for your hard work. It looks good!

>  I had to add "keywords":[ ] to the test_png.json because the keywords were not retrieved by the ExifReader. Do you know why?

I think because png does not support exif: https://stackoverflow.com/questions/9542359/does-png-contain-exif-data-like-jpg
Maybe a newer png would already.

> mocha issue

I'm not sure why it is not running for you. I usually just use my IDE to start the tests:
kép

This would run something like this on windows:
```
"<path1>\nodejs\node.exe" <path2>\PiGallery2\node_modules\mocha\bin\mocha --require ts-node/register --ui bdd --reporter <path3>\WebStorm\ch-0\211.6693.108\plugins\NodeJS\js\mocha-intellij\lib\mochaIntellijReporter.js <path2>\PiGallery2\test\backend\unit\model\threading\MetaDataLoader.spec.ts --grep "^MetadataLoader "
``` 
But running only `mocha test/**/*.js --grep "^MetadataLoader"` also worked for me. 
For issue I can imagen the working dir was not the project , you did not have mocha installed globally, the test files were not trascoded to js

Issue #274, opened 2021-04-26
use Exifreader 3.15.0 to avoid overwriting Rating by RatingPercent


kagahd:
Some Software, as for example XnView, add besides the normal `Rating` which goes from 0 to 5, a `RatingPercent` which goes from 0 to 100. ExifReader overwrites the `Rating` by `RatingPercent`. 
I opened an [issue](https://github.com/mattiasw/ExifReader/issues/148) which has been solved already yesterday, resulting in a new ExifReader version 3.15.0. PiGallery2 should use this new ExifReader version so that the normal `Rating` gets not overwritten by `RatingPercent` in case both exist.
I'll create a pull request with a test to cover this case.

PR #275, opened 2021-04-26
#274 use ExifReader 3.15.0 to avoid overwriting Rating by RatingPercent

kagahd:
I added a test case with a photo having `Rating:3` and `RatingPercent:50`. 
ExifReader before version 3.15.0 will overwrite `Rating` by `RatingPercent` so that PiGallery2 would show 50 instead of 3 as rating.
ExifReader v3.15.0 (and later) solves this problem by introducing a new field `RatingPercent` so that `Rating` wil not be overwritten by `RatingPercent` anymore.
This PR is based on the actual master of PiGallery2 (and not on my PR #270).


bpatrik:
Thank you!
The Metadata loader needs eventually a major refactor (see #277), I just did not have the time for that. (also do no touch if it works :) )

Issue #52, opened 2019-02-18
[Suggestion] download photos


mcdamo:
This is a feature request to download multiple photos in a zip file.

When I share a link with family or friends they want to download the collection of photos easily. This could be an option to download all photos in current folder or an interface to select photos in the thumbnail view and then download only the selected photos.

Thankyou for this app, I'm really impressed with the searching tools.

bpatrik:
Hi,

Thanks for the suggestion. 
I can add the full folder download to the app. Selecting a photo is currently not possible and not super straightforward to implement, I might not do that. Also it will be probably a slow feature if you have like 50+ or even 100+.  I can imagen on weak device (like a raspberry) it can be even like 5-10mins.
Unfortunately I cannot promise any deadline. (Most likely will take up several months.)

Patrik

ALutz273:
This function (download all images as zip) is unfortunately also missing for me and would be very cool
How can you be supported? Unfortunately, I am not a good programmer.

mcdamo:
#278 adds ability to download directory's media in a zip.

bpatrik:
made it support search results too in #1017

PR #278, opened 2021-04-28
Download zip folder

mcdamo:
- adds a config setting to Other/Misc for DownloadZip, defaults to False.
- when enabled, shows a download link on the navigation bar alongside the count of items
- download streams a zip of the current directory's media (photos and videos)

This resolves Issue #52

### DIFF
--- a/src/backend/middlewares/GalleryMWs.ts
+++ b/src/backend/middlewares/GalleryMWs.ts
@@ -72,16 +72,6 @@ export class GalleryMWs {
     }
 
     try {
-      const directory = await ObjectManagers.getInstance()
-        .GalleryManager.listDirectory(directoryName,
-          parseInt(req.query[QueryParams.gallery.knownLastModified] as string, 10),
-          parseInt(req.query[QueryParams.gallery.knownLastScanned] as string, 10));
-
-      if (directory == null) {
-        req.resultPipe = new ContentWrapper(null, null, true);
-        return next();
-      }
-
       res.set('Content-Type', 'application/zip');
       res.set('Content-Disposition', 'attachment; filename=Gallery.zip');
 
--- a/src/backend/middlewares/GalleryMWs.ts
+++ b/src/backend/middlewares/GalleryMWs.ts
@@ -81,12 +81,6 @@ export class GalleryMWs {
         req.resultPipe = new ContentWrapper(null, null, true);
         return next();
       }
-      if (req.session.user.permissions &&
-        req.session.user.permissions.length > 0 &&
-        req.session.user.permissions[0] !== '/*') {
-        (directory as DirectoryDTO).directories = (directory as DirectoryDTO).directories.filter((d): boolean =>
-          UserDTOUtils.isDirectoryAvailable(d, req.session.user.permissions));
-      }
 
       res.set('Content-Type', 'application/zip');
       res.set('Content-Disposition', 'attachment; filename=Gallery.zip');
--- a/src/backend/middlewares/GalleryMWs.ts
+++ b/src/backend/middlewares/GalleryMWs.ts
@@ -16,6 +16,7 @@ import {QueryParams} from '../../common/QueryParams';
 import {VideoProcessing} from '../model/fileprocessing/VideoProcessing';
 import {SearchQueryDTO, SearchQueryTypes} from '../../common/entities/SearchQueryDTO';
 import {LocationLookupException} from '../exceptions/LocationLookupException';
+import {SupportedFormats} from '../../common/SupportedFormats';
 
 
 export class GalleryMWs {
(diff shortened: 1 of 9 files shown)


src/frontend/app/ui/settings/other/other.settings.component.html
```diff
@@ -60,6 +60,13 @@ <h5 class="card-header">
         [ngModel]="states.Client.captionFirstNaming">
       </app-settings-entry>
 
+      <app-settings-entry
+        name="Download Zip"
+        description="Enable download zip of a directory contents"
+        i18n-description i18n-name
+        [ngModel]="states.Client.enableDownloadZip">
```
bpatrik, line 67: Not sure if this works here with [ngModel], you might need [(ngModel)], please test, when hitting save, the change gets saved to config.josn

src/backend/middlewares/GalleryMWs.ts
```diff
@@ -56,6 +57,75 @@ export class GalleryMWs {
...
+      // append photos in selected directory
+      for (const ext of SupportedFormats.WithDots.Photos) {
+        archive.glob(`*${ext}`, {cwd:absoluteDirectoryName, nocase:true});
+      }
+      // append videos in selected directory
+      for (const ext of SupportedFormats.WithDots.Videos) {
+        archive.glob(`*${ext}`, {cwd:absoluteDirectoryName, nocase:true});
+      }
+
+      await archive.finalize(function(err: any) {
```
bpatrik, line 117: Whit out looking into it, this looks odd at the first sight:
You are using a callback function, while also promise return. If this not intentional, that might cause race conditions.

src/backend/middlewares/GalleryMWs.ts
```diff
@@ -56,6 +57,75 @@ export class GalleryMWs {
...
+        console.log('zip ' + archive.pointer() + ' bytes');
+      });
+
+      archive.on('error', function(err: any) {
+        throw err;
+      });
+
+      archive.pipe(res);
+
+      // append photos in selected directory
```
bpatrik, line 108: pls add comment, that this line performs case insesitive match for the ext in `absoluteDirectoryName`

src/backend/middlewares/RenderingMWs.ts
```diff
@@ -71,6 +71,13 @@ export class RenderingMWs {
     return res.sendFile(req.resultPipe, {maxAge: 31536000, dotfiles: 'allow'});
   }
 
+  public static renderZipStream(req: Request, res: Response, next: NextFunction): any {
```
bpatrik, line 74: Having this function is not necessary, if you pipe the result directly into the `res` Object in the ` GalleryMWs.zipDirectory,` function. Also this function currently does not really server a purpose :)

src/backend/middlewares/GalleryMWs.ts
```diff
@@ -56,6 +57,75 @@ export class GalleryMWs {
...
+    try {
+      if ((await fsp.stat(absoluteDirectoryName)).isDirectory() === false) {
+        return next();
+      }
+    } catch (e) {
+      return next();
+    }
+
+    try {
+      const directory = await ObjectManagers.getInstance()
```
bpatrik, line 75: Running `listDirectory here` is not necessary. This code here only checks if the directory is part of the gallery by trying to list it, but the previous middle-ware (`AuthenticationMWs.authorisePath('directory', true)` in the galleryrouter ) has already done that. 
And while this checks here costs a lot (by reading all files in the dear or accessing DB), the previous was O(1) by comparing paths.

src/backend/middlewares/GalleryMWs.ts
```diff
@@ -56,6 +57,75 @@ export class GalleryMWs {
...
+        req.session.user.permissions.length > 0 &&
+        req.session.user.permissions[0] !== '/*') {
+        (directory as DirectoryDTO).directories = (directory as DirectoryDTO).directories.filter((d): boolean =>
+          UserDTOUtils.isDirectoryAvailable(d, req.session.user.permissions));
+      }
+
+      res.set('Content-Type', 'application/zip');
+      res.set('Content-Disposition', 'attachment; filename=Gallery.zip');
+
+      const fs = require('fs');
```
bpatrik, line 94: please add the includes to the top of file, following the `import * as path from 'fs'`; or `import {archiver} from 'archiver'` if it is typescript compatible.

src/backend/middlewares/GalleryMWs.ts
```diff
@@ -56,6 +57,75 @@ export class GalleryMWs {
...
+        return next();
+      }
+      if (req.session.user.permissions &&
+        req.session.user.permissions.length > 0 &&
+        req.session.user.permissions[0] !== '/*') {
+        (directory as DirectoryDTO).directories = (directory as DirectoryDTO).directories.filter((d): boolean =>
+          UserDTOUtils.isDirectoryAvailable(d, req.session.user.permissions));
+      }
+
+      res.set('Content-Type', 'application/zip');
```
bpatrik, line 91: Generally, I tried to separate 'production' middle-wares (like: getting the dir and calling a model to Zip it), from the 'model' that does something (like zips the directory) and from 'rendering' middle-wares (return with the zipped connent).

Unfortunately, all three happens here in the next few lines. 
I cannot really come up, with any solution for separating them, so it should be okay for this case. Just wanted to let you know in case, it case it gives you some ideas.

src/frontend/app/ui/gallery/navigator/navigator.gallery.component.html
```diff
@@ -15,6 +15,9 @@
   </ol>
 
   <div class="right-side">
+    <a *ngIf="config.Client.Other.enableDownloadZip && directory && ItemCount > 0" [href]="'/api/gallery/zip/' + getDirectoryPath()"
```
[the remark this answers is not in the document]
mcdamo: Added urlBase to link

src/frontend/app/ui/settings/other/other.settings.component.html
mcdamo: It works for me

src/backend/middlewares/GalleryMWs.ts
```diff
@@ -56,6 +57,75 @@ export class GalleryMWs {
...
+      const directory = await ObjectManagers.getInstance()
+        .GalleryManager.listDirectory(directoryName,
+          parseInt(req.query[QueryParams.gallery.knownLastModified] as string, 10),
+          parseInt(req.query[QueryParams.gallery.knownLastScanned] as string, 10));
+
+      if (directory == null) {
+        req.resultPipe = new ContentWrapper(null, null, true);
+        return next();
+      }
+      if (req.session.user.permissions &&
```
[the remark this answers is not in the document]
mcdamo: I'll take your word for this and remove it.

src/backend/middlewares/GalleryMWs.ts
mcdamo: Removed listDirectory

src/backend/middlewares/GalleryMWs.ts
mcdamo: Removed callback function.

src/backend/middlewares/GalleryMWs.ts
mcdamo: added comment

src/backend/middlewares/RenderingMWs.ts
mcdamo: Removed renderZipStream

src/backend/middlewares/GalleryMWs.ts
mcdamo: Added import

src/backend/middlewares/GalleryMWs.ts
mcdamo: I understand your design wishes but I don't think there is any benefit to increasing complexity with a generalised Zip rendering MW while it is only used by this specific function.

src/frontend/app/ui/gallery/navigator/navigator.gallery.component.html
```diff
@@ -15,6 +15,9 @@
   </ol>
 
   <div class="right-side">
+    <a *ngIf="config.Client.Other.enableDownloadZip && directory && ItemCount > 0" [href]="getDownloadZipLink()"
```
bpatrik, line 18: Can you also disable it if the app is showing a search result. (Or implement it for search too, but I'm fine it to have it only for folders at this point)

src/backend/middlewares/GalleryMWs.ts
bpatrik: Yes it is fine like this. 
(In the future, the app might need to support search zipping. For that a possible solution would be having a middle-ware that can zip the media from a ContentWrapper object. And then it can use the result GalleryMW listDirectory or search middle-wares. )

bpatrik:
Thank you. It looks good now. Can you please also resolve the inconsistency in search (The button should not be there or should also work for search results), then I can merge it.

src/frontend/app/ui/gallery/navigator/navigator.gallery.component.html
mcdamo: When is it showing the download link for a search result? I have not seen this happen.

bpatrik: Oh indeed the `directory ` catches it, sorry I was not running the code locally.

bpatrik:
Thank you for the contribution and your hard work!

Issue #293, opened 2021-05-08
Faces not indexed if Override keywords is disabled


mcdamo:
Faces are not indexed if _Override keywords_ option is disabled.

This looks like the cause:
https://github.com/bpatrik/pigallery2/blob/cb10227e3c2c1f1895ba4d691115de824e30d6b5/src/backend/model/threading/MetadataLoader.ts#L293-L294

Issue #156, opened 2020-05-14
Support XMP sidecar files


properlypurple:
**Is your feature request related to a problem? Please describe.**
I use Digikam and Darktable both to manage my photos, and I like keeping the metadata in XMP sidecar files, which makes it a lot easier to maintain interoperability, as well as file syncing and versioning for backup. Currently Pigallery2 reads tags directly from the image file, but ignores XMP sidecar files, thus making it not very useful for tags/faces, etc

**Describe the solution you'd like**
When the metadata is loaded, look for and load from xmp sidecar files as well as the image.

Pigallery2 is the best tool I have found yet, to maintain a read-only gallery for my already managed filesystem hierarchy of photos. I already have a syncing and backup system in place, as well as the apps to manage the photos. Support for sidecar files will make this feature complete for me. I'm happy to provide sample images and files for development/testing.

bpatrik:
This is actually a great idea. 

Would you be interested contributing this to the gallery? I can give you some pointers, shouldn't be that complicated. 
If not, I'll live this issue open and once I catch up with my life, I'll implement it.

properlypurple:
Thanks @bpatrik . My javascript is not great, but I'd love to give this a try. I'll set up a dev environment some time this week and will look into this. I'd love any pointers you can give me.

bpatrik:
Hi,
Sorry for the late reply. If you still have the time to do this feature here are the pointers:

dev env:
1) download source
2) `npm install` to get all dependencies
3) `npm run run-dev`  (you can kill the process once it is finished building the fronted as you not going to change that)
4) `npm run build-backend` this only compiles the typescript files, you can also run `tsc -w` if zou have typescript installed globally
4) `npm start` to run the server

Where to change:
The metadata loading is happening in the MetadataLoader.ts file: https://github.com/bpatrik/pigallery2/blob/af012060ec420fabd8938797a38b6210dc40d3a4/src/backend/model/threading/MetadataLoader.ts#L80
I'm reading the beginning of the photo file here: https://github.com/bpatrik/pigallery2/blob/af012060ec420fabd8938797a38b6210dc40d3a4/src/backend/model/threading/MetadataLoader.ts#L84-L85
Then I'm using multiple libraries to parse the metadata (unfortunately there is not a good what that can do everything on its own). I'm using these libs: `ts-exif-parser, ts-node-iptc, exifreader, image-size`.
Please check if any of those support reading the sidecar files, before adding a new dependency to the project. 

[...]

Issue #212, opened 2020-12-27
Incorrect aspect ratio for thumbnail/preview


cbhushan:
**Describe the bug**
Thumbnail/preview has incorrect aspect ratio for one of my images, as shown in screenshot below. The image is also linked below. 

**Photo/video (optional) that causes the bug**
See zip: https://www.dropbox.com/s/ee7vw513vl79nuv/testimg.tar.gz?dl=0 

**Screenshots**
Left thumbnail has correct aspect ratio. Right one (__a20180218_143832_dc8f44e06e4a.jpg) is incorrect. Both images are attached in link above.
![image](https://user-images.githubusercontent.com/3436053/103162601-9be98f00-47c0-11eb-9559-2193fb9d21f9.png)

**Environment (please complete the following information):**
 - Browser: Firefox 84.0 (64-bit)
 - OS: Ubuntu 20.04.1
 
**Used app version**:
- docker-nightly. See screenshot below for version.
![image](https://user-images.githubusercontent.com/3436053/103162679-832da900-47c1-11eb-93c2-54f88f2d5fca.png)

docker-compose.yml:
```
version: '3'
services:
  pigallery-db:
    container_name: pigallery-db
    image: mariadb
    volumes:
      - db-data:/var/lib/mysql
    environment:
      - MYSQL_ROOT_PASSWORD=putsomethinginherethatyouwouldlikeasthepassword
      - MYSQL_USER=pigallery2
      - MYSQL_PASSWORD=pigallery2_pass
      - MYSQL_DATABASE=pigallery2

  pigallery2:
    image: bpatrik/pigallery2:nightly-stretch
    command: sh -c 'bin/wait-for pigallery-db:3306 -- --Server-Database-mysql-host=pigallery-db --Server-Database-mysql-username=pigallery2 --Server-Database-mysql-password=pigallery2_pass --Server-Database-mysql-database=pigallery2'
    container_name: pigallery2
    environment:
      - NODE_ENV=production
    volumes:
      - "/mnt/data/tmp/PiGallery2/config:/app/data/config" # CHANGE ME
      - "db-data:/app/data/db"
      - "/mnt/data/tmp/pic_collection:/app/data/images/pic_collection" # CHANGE ME
      - "/mnt/data/tmp/pic_collection2:/app/data/images/pic_collection2" # CHANGE ME
      - "/mnt/data/tmp/PiGallery2/tmp:/app/data/tmp" # CHANGE ME
    ports:
      - 80:80
[...]

PR #294, opened 2021-05-08
Replace metadata libs with exiftool

### LINKED ISSUE #277: Clean up MetadataLoader.ts
Pigallery2 is using multiple exif, iptc parers (as I could not find one that can do all at once) to get all the metadata for a photo:

1) fs.statSync for silesize, creation date (if i do not find any better from exif)
https://github.com/bpatrik/pigallery2/blob/36ecbcbd3d5d3e659ddf20add456982dcb960d3d/src/backend/model/threading/MetadataLoader.ts#L107

2) `ts-exif-parser` for most of the exif data:  https://github.com/bpatrik/pigallery2/blob/36ecbcbd3d5d3e659ddf20add456982dcb960d3d/src/backend/model/threading/MetadataLoader.ts#L114

3) `image-size` to get the photo dimension if 2) `ts-exif-parser` fails: https://github.com/bpatrik/pigallery2/blob/36ecbcbd3d5d3e659ddf20add456982dcb960d3d/src/backend/model/threading/MetadataLoader.ts#L169
(The image dimension is a must for this app to work properly )

4) `ts-node-iptc` for GPS/location, keywords, caption related info: https://github.com/bpatrik/pigallery2/blob/36ecbcbd3d5d3e659ddf20add456982dcb960d3d/src/backend/model/threading/MetadataLoader.ts#L177

5) `exifreader` for faces and now for orientation as this lib parses that properly https://github.com/bpatrik/pigallery2/blob/36ecbcbd3d5d3e659ddf20add456982dcb960d3d/src/backend/model/threading/MetadataLoader.ts#L206

This class needs a major refactor. Probably. `exifreader` would be able to handle all cases now. Reducing to only one lib, photo indexing would be probably way faster (instead of 4-5 file open, 1 would do the trick)

mcdamo:
Add exiftool and use for photo metadata and sidecars for all media.

- resolves XMP sidecar support #156 
- resolves Faces indexing #293 
- resolves #277

Sidecars are indexed for all media, but UI does not presently display this metadata for **videos**.


bpatrik:
I see that you are changing multiple test cases, that seems to be alarming.

bpatrik:
Please check the failing tests. 
I'll take a deeper look later.

mcdamo:
> Please check the failing tests.
> I'll take a deeper look later.

There is one failing test:
```
  1) MetadataLoader
       should load jpg with edge case exif data
         date_error.JPG:
...
      -  "creationDate": 1620456848859
      +  "creationDate": 0
```
This JPG does not have any valid dates, so metadata is using the file modification date as designed.

I think the test is wrong because it expects an empty date. I don't know how to fix it because the file modification date varies depending on when git checked out this file.

src/backend/model/threading/MetadataLoader.ts
```diff
@@ -86,232 +101,269 @@ export class MetadataLoader {
...
+      Logger.debug(LOG_TAG, 'Error parsing exif', fullPath, err);
+      throw err;
+    }
+    return metadata;
+  }
 
-            } catch (err) {
-              // Logger.debug(LOG_TAG, 'Error parsing iptc data', fullPath, err);
-            }
+  private static decodeExifCameraData(exif: Tags, cameraData: CameraMetadata): any
```
bpatrik, line 193: Can you please avoid using `any` if possible (only flagging it here, but applies to all the decode functions).
`any` makes code harder to read (now one should check the implementation to see what it will return) and also prevents the compiler to warn you if a type is not correct.

src/backend/model/threading/MetadataLoader.ts
```diff
@@ -86,232 +101,269 @@ export class MetadataLoader {
...
-                      regionRoot.value['rdf:Description'].value &&
-                      regionRoot.value['rdf:Description'].value['mwg-rs:Area']) {
+  private static decodeExifCreationDate(exif: Tags, creationDate: any): any
+  {
+    if (exif.CreateDate instanceof ExifDateTime || exif.DateTimeOriginal instanceof ExifDateTime || exif.ModifyDate instanceof ExifDateTime) {
+      const myDate = (exif.DateTimeOriginal instanceof ExifDateTime && exif.DateTimeOriginal ||
+        exif.CreateDate instanceof ExifDateTime && exif.CreateDate ||
+        exif.ModifyDate instanceof ExifDateTime && exif.ModifyDate).toDateTime();
+      // get unix time in original timezone
+      creationDate = myDate.toMillis() + (myDate.offset * 60 * 1000);
```
bpatrik, line 275: The test with the datetime fails probably because of this line. It overrides the value in line 114: `metadata.creationDate = stat.mtime.getTime();`

I know this a half solution. Basically what is happening here is if a photo does not have a proper creation date, it falls back to the file creation/modification date. This is not ideal, but at least if you open a folder and list the photos by date, it would give some approximation for ordering. (A better solution would be here to actually flag this date and show on the UI that this only an inferred value and probably broken. A distant future, I might do it)

Once you fix it, the test will pass because it would use the time it got created/modified within git and that is a fixed value.

src/backend/model/threading/MetadataLoader.ts
```diff
@@ -86,232 +101,269 @@ export class MetadataLoader {
...
-                      const regionBox = region.value['mwg-rs:Area'].attributes;
+  private static decodeExifSize(exif: Tags, size: any, orientation: any = null): any
+  {
+    if (exif.ImageWidth) {
+      size = {width: exif.ImageWidth, height: exif.ImageHeight};
+    } else if (exif.RelatedImageWidth && exif.RelatedImageHeight) {
+      size = {width: exif.RelatedImageWidth, height: exif.RelatedImageHeight};
+    }
+    if (orientation !== null && orientation !== undefined) {
+      if (orientation > OrientationTypes.BOTTOM_LEFT) {
```
bpatrik, line 288: You can hard code the value `4` (instead of BOTTOM_LEFT) and remove the `ts-exif-parser`.  Maybe add a comment what that means.

Here are the orientation type: https://github.com/bpatrik/ts-exif-parser/blob/master/lib/ExifData.ts

test/backend/assets/orientation/broken_orientation_exif2.json
```diff
@@ -9,9 +9,9 @@
   },
   "creationDate": 1185452882000,
   "fileSize": 124036,
-  "orientation": 1,
+  "orientation": 8,
```
bpatrik, line 12: This is a tricky one:

There is a huge thread about the orientation issue: #212
I would like to avoid restarting it. 

The TL;DR:

The `size -> {height,width}` should contain the post rotation size. The `orientation` tag is not used anymore anywhere (this is on my, I was lazy to clean it up properly).

How it actually works:
`size` is used to render the UI grid, the grid cells itself and for the preview (lightbox) sizing.
Thumbnails and generated photos are done with sharp that understands and reads orientation tag to transcode a photo. Browsers now understand orientation tag and will render the photo properly. 

You can test these visually by running the gallery with all 9 stock orientation photos and the 3 problematic photos. (you can find all of them in #212 )

src/backend/model/threading/MetadataLoader.ts
mcdamo: I understand that it is using the file modification date when EXIF is not available, and agree this is the best practice.

Git does not store the file timestamp, when you clone or checkout files they are given a new timestamp. If you try this you will see. The existing test was erroneous because it expected createdDate=0 instead of the file timestamp, and the erroneous test was passing because the existing code had a bug that would set the creationDate=0 if there was no dates in the EXIF data.

package.json
```diff
@@ -40,19 +40,17 @@
...
     "express-unless": "0.5.0",
     "fluent-ffmpeg": "2.1.2",
-    "image-size": "0.9.7",
     "jimp": "0.16.1",
     "locale": "0.1.0",
     "node-geocoder": "3.27.0",
     "reflect-metadata": "0.1.13",
     "sharp": "0.23.4",
     "sqlite3": "5.0.0",
     "ts-exif-parser": "0.2.1",
```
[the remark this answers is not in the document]
mcdamo: Removed

package.json
```diff
@@ -94,6 +92,7 @@
     "@types/jsonwebtoken": "8.5.1",
     "@types/leaflet": "1.7.0",
     "@types/leaflet.markercluster": "1.4.4",
+    "@types/luxon": "^1.26.5",
```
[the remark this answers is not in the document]
mcdamo: This appears to be required for exiftool-vendored

src/backend/model/threading/MetadataLoader.ts
mcdamo: Fixed

test/backend/assets/orientation/broken_orientation_exif2.json
mcdamo: After trying many different options I think this specific image is broken in such a way that it is not useful as a test, many programs I tested do not agree if this is a portrait or landscape image and the data from different EXIF tools is inconsistent.

bpatrik: At some extend you are right. The issue with this photo is that:
"
Apparently both the photo and the embedded thumbnail can both have orientations. It is also possible that their orientation is not the same. In case of roof.jpg The thumbnail orientation is 8, while the photo orientation is 1. This lib was returning 8 not specifying if it belonged to the photo or the thumbnail.
" #212

This is a very weird edge case. We can argue that we do not support it in the app, but it "just works" in Win explorer and other photo viewers, so I think we cannot discard it either.

bpatrik: Had a quick look with some problematic photos. (Without properly understanding it). I used `exiftool(-a -k -u -g1).exe`.

With the roof photo I saw that i has:

```
---- IFD0 ----
Make                            : Panasonic
Camera Model Name               : DMC-LZ7
X Resolution                    : 72
Y Resolution                    : 72
Resolution Unit                 : inches
Software                        : Ver.1.0
Modify Date                     : 2007:07:26 12:30:54
Y Cb Cr Positioning             : Co-sited
Custom Rendered                 : Normal
Exposure Mode                   : Auto
White Balance                   : Auto
Digital Zoom Ratio              : 0
Focal Length In 35mm Format     : 37 mm
Scene Capture Type              : Standard
Gain Control                    : None
Contrast                        : Normal
Saturation                      : Normal
Sharpness                       : Normal
---- IFD1 ----
Compression                     : JPEG (old-style)
Orientation                     : Rotate 270 CW
X Resolution                    : 72
Y Resolution                    : 72
Resolution Unit                 : inches
Thumbnail Offset                : 7808
Thumbnail Length                : 8874
Y Cb Cr Positioning             : Co-sited
Thumbnail Image                 : (Binary data 8874 bytes, use -b option to extract)
```

while some other portrait photo (Like an Iphone one) :
```
---- IFD0 ----
Make                            : Apple
Camera Model Name               : iPhone 8
Orientation                     : Rotate 90 CW
X Resolution                    : 72
Y Resolution                    : 72
Resolution Unit                 : inches
Software                        : 11.4.1
Modify Date                     : 2018:08:23 19:37:12
Y Cb Cr Positioning             : Centered
```

What you can see:
the roof photo has both IFD0 and IFD1, while the IPhone only IFD0.
For roof IFD0 does not contain orientation, while the iphone one does.
For roof IFD1 contains orientation.

I **think** (without properly understanding it):
* IFD0 corresponds to the main image, while IFD1 to the embedded thumbnail.
* roof photo is properly rotated, therefore there is no orientation tag in IFD0
* Iphone needs rotation therefor there is an orientation tag in IFD0
* Iphone does not have embedded thumbnail, therefore there is no IFD1 there

mcdamo: Interesting, if thats the case that IFD1 relates to thumbnail then we can exclude reading that page of EXIF and that will resolve the result of this test as it was originally: 6fed317

src/backend/model/threading/MetadataLoader.ts
```diff
@@ -86,232 +101,269 @@ export class MetadataLoader {
...
+        // search for sidecar and merge metadata
+        const fullPathWithoutExt = path.parse(fullPath).name;
+        const sidecarPaths = [
+          fullPath + '.xmp',
+          fullPath + '.XMP',
+          fullPathWithoutExt + '.xmp',
+          fullPathWithoutExt + '.XMP',
+        ];
+        for (const sidecarPath of sidecarPaths) {
+          if (fs.existsSync(sidecarPath)) {
```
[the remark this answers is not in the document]
mcdamo: I can't test on a RPi, but generally the indexing with this update is about twice as slow for me.

Your benchmark example doesn't add up, maybe wrong units?
> Checking if a file exist takes about 0.09**ms** on an RPi:
> This part of the code on a gallery with 1000 photos: would use 0.09 * 1000 * 4=3.6s.

0.09ms * 1000 * 4 is 360**ms**, or 0.36**s**

The block that tests for existing `.xmp` files did not make any material difference to the indexing speed for me.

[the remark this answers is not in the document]
mcdamo: If exifr can support XMP sidecars that would be ideal, but considering the scope of XMP it looks like it could be a massive undertaking. We can wait some time for exifr to decide if they will add support.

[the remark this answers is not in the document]
bpatrik: It seems to happen soon. exifr already supported xmp, just not sidecars, but that is an easy update (calling the right function for the file)

bpatrik:
exifr now support sidecars: https://github.com/MikeKovarik/exifr/issues/62

mcdamo:
> exifr now support sidecars: [MikeKovarik/exifr#62](https://github.com/MikeKovarik/exifr/issues/62)

exifr looked promising but sadly it has not had any activity from the maintainer in the past 12 months.

bpatrik:
yeah. I tried to migrate to it a few months ago and it was still missing some features. That is unfortunate.

bpatrik:
I ended up merging some changes in this area since after a long time I did not get there to refactor the metadata loader.

i will reject this PR just for now.

PR #303, opened 2021-05-19
Python version unavailable for alpine 3.13

martadinata666:
As python in alpine 3.11 is version 2.x.x and python in alpine 3.13 have two versions ```python3``` and ```python2```.


martadinata666:
I'm trying as close as possible to 3.11 deps that using python2, but we can use python3 of course. 👍🏼

bpatrik:
Thanks, lets try it.

PR #331, opened 2021-06-23
Serve custom icon if it exists in /config

niawag:
Add a check to serve a custom icon if it exists in the config folder.

I hope I got the logic behind app.get function, I am no js expert...

### DIFF
--- a/src/backend/routes/PublicRouter.ts
+++ b/src/backend/routes/PublicRouter.ts
@@ -137,7 +137,16 @@ export class PublicRouter {
       setLocale,
       AuthenticationMWs.normalizePathParam('file'),
       renderFile());
-
+    app.get('/assets/icon.png',
+      setLocale,
+      (req: Request, res: Response, next: NextFunction)=>{
+        if (fs.existsSync('../../data/config/icon.png')) {
+          return res.sendFile('../../data/config/icon.png');//not sure about what the path to /config is
+        }
+        else {
+          return res.sendFile('/assets/icon.png');
+        }
+      },// I don't know if "AuthenticationMWs.normalizePathParam('file')" is needed here
   }
 
 }


bpatrik:
It looks good. The test fail because of the `,` at line 150. Also can you also make it work for the inverse icon?

But the best would probably be to have a `/config/assets_overrid` folder that can be used to override any files in the app's assests folder. Can could have been done somewhere here: https://github.com/bpatrik/pigallery2/blob/3e33d8ab895cf532ba3d79a400478274740da981/src/backend/routes/PublicRouter.ts#L135

src/backend/routes/PublicRouter.ts
```diff
@@ -137,7 +137,16 @@ export class PublicRouter {
...
       renderFile());
-
+    app.get('/assets/icon.png',
+      setLocale,
+      (req: Request, res: Response, next: NextFunction)=>{
+        if (fs.existsSync('../../data/config/icon.png')) {
+          return res.sendFile('../../data/config/icon.png');//not sure about what the path to /config is
+        }
+        else {
+          return res.sendFile('/assets/icon.png');
```
bpatrik, line 147: this path is probably wrong. 

This is how you get the correct path within the app: https://github.com/bpatrik/pigallery2/blob/3e33d8ab895cf532ba3d79a400478274740da981/src/backend/routes/PublicRouter.ts#L122-L130

src/backend/routes/PublicRouter.ts
```diff
@@ -137,7 +137,16 @@ export class PublicRouter {
       setLocale,
       AuthenticationMWs.normalizePathParam('file'),
       renderFile());
-
+    app.get('/assets/icon.png',
+      setLocale,
+      (req: Request, res: Response, next: NextFunction)=>{
+        if (fs.existsSync('../../data/config/icon.png')) {
+          return res.sendFile('../../data/config/icon.png');//not sure about what the path to /config is
```
bpatrik, line 144: huh, this is actually a more complicated question than I first thought. 

The config location is set here: https://github.com/bpatrik/pigallery2/blob/3e33d8ab895cf532ba3d79a400478274740da981/src/common/config/private/Config.ts#L14

But it is overwritten if it runs in the docker here: https://github.com/bpatrik/pigallery2/blob/3e33d8ab895cf532ba3d79a400478274740da981/docker/debian-buster/Dockerfile.build#L38

I'm not sure if you can ask the config object where it lives, need to look. (or implement that function)

bpatrik:
I'm going to close this as the latest version supports custom svg icons

PR #339, opened 2021-07-19
Fix whitespace between Navbar and breadcrumbs

ashwin-pc:
There is a whitespace between the Navbar and breadcrumb/sort section. The change here is to remove the margin style that causes the white space.

Screenshot:
Screenshot 2021-07-18 at 9 13 16 PM

### DIFF
--- a/src/frontend/app/ui/frame/frame.component.css
+++ b/src/frontend/app/ui/frame/frame.component.css
@@ -2,10 +2,6 @@ nav {
   z-index: 999;
 }
 
-.navbar {
-  margin-bottom: 20px;
-}
-
 ng2-slim-loading-bar {
   /* display: block; */
   z-index: 999999;


bpatrik:
I do not find that to be a bug. Why should they be together?

ashwin-pc:
My apologies If i called this a bug somewhere, it was more of a fit and finish thing. The distinct dark grey nav bar with the light grey breadcrumbs bar having a 20px space in between was throwing me off for a long time, especially on smaller screens so i thought i could help out and make it a lil better. 

My suggestions based on my previous UX experience and a quick look at some of the most popular design systems out there (e.g. with subnav: [Bootstraps homepage](https://getbootstrap.com/docs/5.0/getting-started/introduction/), [AWS](https://aws.amazon.com/machine-learning/?hp=tile&tile=solutions)). The whitespace definitely make a lot more sense if the subnav didnt already have a distinct light grey color to separate it visually from the content around it(e.g. [Github](https://github.com/marketplace?type=apps)).

Is there a particular reason for that margin that im not seeing? The light grey breadcrumb section is distinct enough that the space seems irrelevant. It is also padded well enough that removing the margin makes the fit and finish just a little more polished.

Dont get me wrong, this app is designed beautifully. Its just this whitespace that stood out like a sore thumb compared to the rest of the app.

bpatrik:
Alright you convinced me, I actually like it. Thanks for the suggestion!

Issue #341, opened 2021-07-19
Gallery sorting issue


antegs:
Having some problems with sorting starting with version 1.8.5. (I think it might be related to PR #230.)

I have 5 images in a folder. When I sort by "ascending name" the order of the images should be:
2021.01.01-01.01.01.jpg
2021.01.01-01.01.02.jpg
2021.01.01-01.01.50.jpg
2021.01.01-01.01.51.jpg
2021.01.01-01.01.52.jpg

But instead the order is:
2021.01.01-01.01.50.jpg
2021.01.01-01.01.51.jpg
2021.01.01-01.01.52.jpg
2021.01.01-01.01.01.jpg
2021.01.01-01.01.02.jpg

**Environment**:
 - OS: docker on linux, browser on linux
 - Browser: Firefox

**Used app version**:
- docker alpine
- release 1.8.5 and above (works on 1.8.2)

bpatrik:
Indeed that `natural-orderby` does not work as expected.
It was added in #230 and #224.


Had a quick look, and `String.prototype.localeCompare` should do the job:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare
https://stackoverflow.com/questions/2802341/javascript-natural-sort-of-alphanumerical-strings

With something like that:
```
private collator = new Intl.Collator(undefined, {numeric: true});
....
this.media.sort((a: PhotoDTO, b: PhotoDTO) => this.collator(a.name, b.name));
```
It needs to be replaced in 2 places: photo and directory sorting. `natural-orderby` should be also removed from the package.json

Let me know if you have the capacity creating a PR to fix this, otherwise I can do it at some point.

antegs:
Good suggestion to use `Intl.Collator`. I created #342 to fix this.

bpatrik:
If I understand correctly this is fixed. Closing it. Feel free to reopen if you think otherwise.

PR #342, opened 2021-07-20
Fix gallery sorting #341

antegs:
As mentioned in the issue, sorting by "ascending name" does not work as expected. This fix removes `natural-orderby` and uses `Intl.Collator` in its place.


bpatrik:
Thank you!

PR #344, opened 2021-07-23
fix encoding non-ascii characters in detected face names

ajgon:
This is a fix [which was also applied in exif reader](https://github.com/mattiasw/ExifReader/pull/47/files#diff-d65f0e04e026c207d67b8b88660ba72716c0e5363af90c7355ea04d1b9b256d6R120) to properly handle UTF-8 characters in face names. The problem only occurs in `exiftool` generated metadata as atttributes are accesed by raw values there (the "lightroom" version of face meta is fine).

Here is an example photo to illustrate the problem: https://transfer.sh/15gTYWl/IMG_5910.jpg

Without patch:
wrong

With patch:
correct


bpatrik:
Thank you!

PR #345, opened 2021-07-23
use local assets instead of CDN

ajgon:
This one falls into "nice to have" category, so please close it if you find it irrelevant.

The idea is, to not to rely on any external CDN (especially for asset this small), as they're usually set tracking cookies, log IPs and do some other nasty stuff. I think, that for the sake of privacy of the user, it's better to serve this locally. What do you think?

### DIFF
--- a/src/frontend/assets/bootstrap-switch.css
+++ b/src/frontend/assets/bootstrap-switch.css
@@ -1,0 +1,187 @@
+/**
+  * bootstrap-switch - Turn checkboxes and radio buttons into toggle switches.
+  *
+  * @version v3.3.4
+  * @homepage https://bttstrp.github.io/bootstrap-switch
+  * @author Mattia Larentis <mattia@larentis.eu> (http://larentis.eu)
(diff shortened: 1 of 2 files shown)


bpatrik:
External css should rater go to the `angular.json`, like: https://github.com/bpatrik/pigallery2/blob/d2d1472763aaa97730f695e1f4ffecb760aab454/angular.json#L74-L81

I'm not sure if the button support its (i.e.: provides a css within the node module). If not, we probably need to replace it.

Also the APP can never get fully "offline" as there is reverse geocoding in search that uses same API.

ajgon:
Unfortunaltely I'm not a JS dev, so if something else needs to be done, to make this work - I won't be a big help here :/ I did this monkey-patch in my container and it worked though :)

Regarding geocoding - I'm not sure what's going on there, but I'm running pigallery2 in a container, with Internet access cut off - and didn't notice any issues. Probably I'm not using said feature.

Issue #220, opened 2021-01-10
How to change admin password?


thve25:
First of all, I'm really happy with this project, its great!

Only one issue, I see no way to change the default admin password. 
I use docker-compose.

Any suggestions?

Thanks,
Thomas

bpatrik:
Hi,

You need to log in with the original `admin` user . Create a new user with Admin rights. Login with that user and delete the original `admin` user.

Patrik

thve25:
Thanks, that makes sense. Sorry for the stupid question

bpatrik:
Created a Config setting where everyone can list the user that they want to enforce in the app. (You also set it through env variables) It should be secure as if you set the password in the config.json, the app will write back the salted, encrypted version and delete the clear text.

bpatrik:
Updated the MANPAGE.md in 58e1857 that shows the usage of the config.

yllekz:
Could you feasibly update the instructions to state that a new user needs to be created + the old "admin" user must be deleted? I also got super confused by this, initially just looking for a "change password" prompt.

stenrock:
I created a new user with Admin rights and delete the Admin user but still get the warning "(1) Using default admin user!
"You are using the default admin/admin user/password, please change or remove it."

I am puzzled.

san-jupitero:
> I created a new user with Admin rights and delete the Admin user but still get the warning "(1) Using default admin user! "You are using the default admin/admin user/password, please change or remove it."
> 
> I am puzzled.

same issue, have no clue how it can be fixed

Gooygeek:
I was having the same issue. Created a new user with admin rights, deleted the old admin user, still kept getting the warning popup.

The solution was a tale as old as computing... Turn it off and on again.

Restarting the container solved the issue.

PR #393, opened 2021-11-21
Allow admin user/password through environment variables

joanbm:
Allows specifying the default admin user / password through the `PI_ADMIN_USER` / `PI_ADMIN_PASSWORD` (or `PI_ADMIN_USER_FILE` / `PI_ADMIN_PASSWORD_FILE`) environment variables, and update the documentation to strongly recommend changing the default password.
(The reference to the possibility to remove the default admin user has been removed, as it allows the possibility of an insecure setup if the database were to be recreated).

Fixes: #220

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,9 @@ npm start
 To configure it, run `PiGallery2` first to create `config.json` file, then edit it and restart.
 The app has a nice UI for settings, you may use that too. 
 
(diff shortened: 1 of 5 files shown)


bpatrik:
Sorry for the late reply. 

Thank you for the contributions, but I found a better place fort his through the config. see e956199 and my comment in #220

Issue #409, opened 2022-01-18
Negative EXIF ratings cause DB malfunction


desertwitch:
Negative EXIF ratings such as -1, which are used by few rare clients for marking images for deletion, cause folder indexing queries to fail because _tinyint_ database field does not tolerate negative values, only positive values from 0 to 255. As part of this problem the image with negative EXIF rating and all following images of that folder are not indexed into the DB.

**driverError: Error: ER_WARN_DATA_OUT_OF_RANGE: Out of range value for column 'metadataRating' at row 1**

- Solution A: Change field type in database from _tinyint_ (1 byte) to _smallint_ (2 byte) to allow negative ratings too and stay true to actual rating of image.
- Solution B: Always override negative ratings to 0 because negative ratings are useless, not staying true to actual rating of image - see my pull request.

bpatrik:
I prefer solution B (min value should be 0) as client only works with value
between 0 and 5

P

-- Sorry for being brief, sent from my phone.

Veronika ***@***.***> ezt írta (időpont: 2022. jan. 18., K
12:53):

desertwitch:
fixed #410

PR #410, opened 2022-01-18
EXIF rating

desertwitch:
see #409 for proposed solutions regarding "No negative EXIF ratings"

### DIFF
--- a/src/backend/model/database/sql/GalleryManager.ts
+++ b/src/backend/model/database/sql/GalleryManager.ts
@@ -222,7 +222,7 @@ export class GalleryManager implements IGalleryManager, ISQLGalleryManager {
    */
   public async fillPreviewForSubDir(connection: Connection, dir: SubDirectoryDTO): Promise<void> {
 
-    if (!dir.preview || !dir.validPreview) {
+    if (!dir.validPreview) {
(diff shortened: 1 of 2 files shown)


src/backend/model/database/sql/GalleryManager.ts
```diff
@@ -222,7 +222,7 @@ export class GalleryManager implements IGalleryManager, ISQLGalleryManager {
    */
   public async fillPreviewForSubDir(connection: Connection, dir: SubDirectoryDTO): Promise<void> {
 
-    if (!dir.validPreview) {
+    if (!dir.preview || !dir.validPreview) {
```
bpatrik, line 225: It's a valid state that preview does not exist and validpreview is true. It means that the assignment code run but did not return a result. Like when a directory is empty. I.e.: not checking the preview here is a feature not a bug.

bpatrik (changes requested):

src/backend/model/database/sql/GalleryManager.ts
desertwitch: Ok, weirdly this fixes the bug described in #411, probably the #411 issue is in another place then.

bpatrik:
Thanks!

PR #456, opened 2022-03-11
Gpx wpt mappoints with infinite loop fix

zigmhount:
This one is the good one :) Sorry for the confusion!


bpatrik:
haha. thanks :)

PR #480, opened 2022-04-28
Use UTC for displaying image timestamps

### LINKED ISSUE #469: Image date displayed in incorrect timezone
Image date is adjusted to the user's browser timezone.

For example if image was taken at 10:30 and the browser is in UTC+10 then date will display as 20:30. This was previously [raised as question here](https://github.com/bpatrik/pigallery2/discussions/360)

Image datetime metadata does not have timezone information so should be displayed without any time adjustments.

mcdamo:
Fixes #469


bpatrik:
Hi,

Thank you for the PR. (Let's hope it covers all the edge cases)

PR #485, opened 2022-05-06
Typo: replace "an other" with "another"

jacroe:
Replace instances of "an other" with "another" throughout the project.

This also touches translation files, which, I don't know if that causes negative effects in tooling.


bpatrik:
Thank you!

Issue #22, opened 2018-12-05
Support IPTC 120 - Caption / Abstract


natrius:
Search for that tag and if its present show it a nice way or show it via key or button. That way i could add information about a picture there and its present, no matter on which device. 
That way there is also no need for a database-field or something like that.

natrius:
Sorry, i forgot to add the links (even if you have already implemented it) http://phpgraphy.sourceforge.net/manual/latest/apas04.html and https://sno.phy.queensu.ca/~phil/exiftool/TagNames/IPTC.html 

Is there one picture in the demo where i can see that implementation?

bpatrik:
No problem, I knew this IPTC tag.
Here is an example: https://pigallery2.herokuapp.com/gallery/?p=IMG_5910.jpg
The caption is visible at the top left corner.

I'm not sure about this version yet. I don't want to pollute the design with additional text and I find the file names more useful on the image tiles, so I only put it to the preview window.

natrius:
I like it that way actually. One thought, some keystrokes are recognized right now (ie. right-arrow, left-arrow), maybe add one "i" for "Information" and "c" for "Caption". 
I tend to use date-time.jpg for my images and seldom change the filename itself. But i might consider it. The nice thing about caption is, it could be a longer description actually. Thats why it is nice sometimes. 

And, make it as an opt-in via settings? That would be another idea. So its default clean and if someone would like to see the captions, its possible. 

Thanks for your work! :)

bpatrik:
I've added the shortcut key for `f` fullscreen , `i` info panel, `c` show controls (caption is part of the control panel)
There is now also a config called (Client->)Other->CaptionFirstNaming (default false). If you change it to true, the captions will be shown in the grid view instead of the filename. 

your welcome :)

FoxxMD:
This is awesome thank you <3

PR #486, opened 2022-05-06
Added code to render title (object_name) instead of filename.

Scott8586:
Addressing Feature request: https://github.com/bpatrik/pigallery2/issues/487

I've added IPTC "title" (object_name) as an option to replace filename, in the same way caption can be used.
My workflow comes almost exclusively from Lightroom, and Title is often what I use as a short phrase for the photo.
(just seems more natural in the Lightroom metadata panel).

I don't know how to create translations, so that part is missing.

(sorry about the misfire on the first request, there was a critical error that I missed in visual testing, and this should be cleaner).

This is my first/second pull request to a public project, and I'm learning node on the fly - so a careful review is warranted.  Thanks for your time!

### DIFF
--- a/benchmark/BenchmarkRunner.ts
+++ b/benchmark/BenchmarkRunner.ts
@@ -164,6 +164,7 @@ export class BenchmarkRunner {
 
       any_text: 'any-text',
       keyword: 'keyword',
+      title: 'title',
       caption: 'caption',
       directory: 'directory',
       file_name: 'file-name',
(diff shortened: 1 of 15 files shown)


bpatrik:
Thank you!

Generally it looks good to me. Can you explain what is the expected with the captionfirst/titlefirst naming?

Originally there was a flag that show caption instead of file names in the galley (on the small photos).
Original FR: #22
How will that work with your change?

bpatrik:
Also please fix the tests. The new metadata tag breaks some of the test: 
kép
More details here: https://github.com/bpatrik/pigallery2/runs/6434869538?check_suite_focus=true

bpatrik:
Rejecting this PR due to inactivity

Issue #498, opened 2022-06-08
Failure to request thumbnails properly when certain special characters are being used (ie, URL escaping is missing)


AndGem:
**Describe the bug**

If a filename contains a question mark (`?`) pigallery2 will not be able to display the picture. For example if there are identical and valid jpg files `main_page.jpg` and `main_page?.jpg` then, the former will be shown correctly, while the latter will not be shown.

**Photo/video (optional) that causes the bug**

- independent of picture

**Screenshots (optional)**

<img width="1440" alt="Bildschirmfoto 2022-06-08 um 21 32 38" src="https://user-images.githubusercontent.com/13299627/172702081-7b1a4e59-0d0a-4e88-be48-f491670ca52d.png">

^ illustration

**Server logs (optional)**

```
6/8/2022, 7:38:05 PM[VERBS] GET /api/gallery/content/test/main_page?.jpg/thumbnail/480 404 2ms
```

Note that requesting manually with `%3F` instead of `?` works :

```
6/8/2022, 7:39:22 PM[VERBS] GET /api/gallery/content/test/main_page%3F.jpg/thumbnail/480 200 59ms
```

**Environment (please complete the following information):**
 - OS: linux server, Ubuntu 20.04
 - Browser macOS, Chrome

**Used app version**:
Using Docker, but I'm struggling to get the version out. It is the current `latest`. If you advice how to get the version number, I'll extract it. Until then, I found this in the logs:

```
[Typeconfig] Processing cli and ENV inputs: {
	"PATH": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
	"HOSTNAME": "2db19b8a737f",
	"NODE_ENV": "debug",
	"NODE_VERSION": "14.19.0",
	"YARN_VERSION": "1.22.17",
	"PI_DOCKER": true,
	"HOME": "/root",
	"VIPS_MIN_STACK_SIZE": "2m",
	"VIPSHOME": "/target"
}
```

daedric7:
I'm still getting this issue, in my case the filename has multiple question marks and spaces around them.

Am i alone here ?

bpatrik:
Is this still an issue with the latest nightly?
#521 seems to solve the issue.

tilmanginzel:
@bpatrik Yep that fixed it, I have tested it with latest master branch. Thumbnail URLs for files with `?` characters are encoded just fine for me.

PR #521, opened 2022-07-20
encoding of question mark character for media path - fixes issue #498

AndGem:
This PR tries to close issue #498 

I've tested the changes locally and they seem to work for question mark characters both in image and video files.

Feedback very welcome.


bpatrik:
thank you

PR #526, opened 2022-08-19
Upgrade ffprobe-static to support darwin arm64

lijunle:
See https://github.com/joshwnj/ffprobe-static/pull/17


bpatrik:
Thanks. Although you do not need it. It can use the system's ffprobe.

PR #529, opened 2022-08-26
Added one useful resource about installing node.js

Vish1161:
Hey, I have added a helpful reference link for installing node.js in the node.js section that I think will best add to your content and give your readers a more diverse understanding of the topic. I hope you will like this. Thank you.


bpatrik:
Thank you!

Issue #535, opened 2022-09-02
Exposure Value in Info-Panel will Overrun rest of Panel


Sean-T-Moore:
**Exposure Value in Info-Panel will overrun rest of Panel**

The code currently takes the reciprocal of the exif exposure value to display as a 1/x fraction. But there is no attempt to constrain the number of decimal places shown, so the value can overrun the focal length value that is next in the panel.

For example, if the exposure value is: 0.002824858757062147
Then the MetadataLoader will store this as:  0.0028 and the panel display code will show this as 1/x like so:
1/357.14285714285717s

![exposure-value-format-problem](https://user-images.githubusercontent.com/6337432/188034270-d345e0fc-446e-4e2f-b68d-f04dc09c9c92.png)

exiftool and other exif viewers will display this as 1/354. 

I have created a branch to fix the precision of the exposure value and to display the fraction without the offending extraneous decimal digits causing the display overrun.

I'm ready to submit a PR if you would like to incorporate this fix.

bpatrik:
To my understanding this is fixed. Closing this. Feel free to reopen if you think otherwise.
Also tanks for flagging it :)

Issue #437, opened 2022-02-26
Improve listing/searcing performance


bpatrik:
Opening directory with a lot (2000+) of photos or running a big search query takes too long.

The issue is two fold:
 1. server side processing
 2, data download time. 

It turns out that 2) is more dominant that 1) if one lists a lot of photos.

Searching for `.` (i.e listing my whole gallery ~60k items), downloaded a 33.3MB json in 1.5min, which is 90x longer than I would prefer it. 
![kép](https://user-images.githubusercontent.com/7512182/155852039-7deab15a-efe7-4dbf-a025-15e28a7e002e.png)

Response header does not show gzipping, but I run into this before and I think the OS unzips and removes it:
![kép](https://user-images.githubusercontent.com/7512182/155852116-54e07669-6c7b-4261-937a-527d541b63e8.png)

I see a few solutions to this:
 1. Save on metadata.
     * I can save some stuff here and there, like no need to send the id down, but that would be just packet change. I want to reduce the size by 50-80%.
 2. Do some json tricks
    * like renaming `name` to `n` in the json object to save on the string lengths. I do not think that would do any help. This is what gzip is for.
  3. trying some other data format. I was looking at [protobuf](https://auth0.com/blog/beating-json-performance-with-protobuf/). Looks promising, but not sure if it worth the investment,

**Blocked by:**
* #299

bpatrik:
Did some analysis.

I found that doing the following modification on the sent data this much can be saved:


| action                                                                                      | data size | additive saving  | additivesaving  [%] |
|---------------------------------------------------------------------------------------------|-----------|-------------------|------------|
| Original data                                                                               | 33.55MB   | N/A               | N/A        |
| Remove DB ids                                                                               | 32.19MB   | 1.36MB            | 4.060%     |
| remove not used orientation                                                                 | 31.17MB   | 2.38MB            | 7.100%     |
| reduce float precision (eg.: gps coordinates do not need 10 dec. precision)                 | 30.15MB   | 3.40MB            | 10.143%    |
| replace readyThumnails field with a bitmap and send missingThumbnail rather                 | 28.77MB   | 4.78MB            | 14.253%    |
| Use list instead of object for face region bounding box                                     | 27.96MB   | 5.60MB            | 16.681%    |
| Mapify directory: Do not send directories with each photo, rather group them                | 23.77MB   | 9.79MB            | 29.168%    |
[...]

PR #536, opened 2022-09-04
Fix problem with display of Exposure Value

Sean-T-Moore:
This PR will fix issue #535


bpatrik:
Thank you!

It will reduce the listing performance a bit #437
, but that's fine.

Issue #534, opened 2022-09-01
Job error: TypeError: config.sizes is not iterable when manually run the thumbnail generation


hajo62:
docker image [v1.9.3] on ubuntu 22.04 on RaspberryPi 4

When I manually run the thumbnail generation, I get the error: `Error: Job error: TypeError: config.sizes is not iterable`.

<img width="624" alt="image" src="https://user-images.githubusercontent.com/15800980/188020521-9024cafe-c47b-4ace-8e79-5855e8f29501.png">

When I start the job in the indexing section, it seems to run:
<img width="1281" alt="image" src="https://user-images.githubusercontent.com/15800980/188021109-9842d3b4-6ca6-48a2-977d-1f2e735948a9.png">

tilmanginzel:
Seems to be a duplicate of #458 (possible fix is mentioned there)

bpatrik:
To my understanding this is fixed. Closing it. Feel free to reopen if you think otherwise.

Issue #458, opened 2022-03-16
Regenerate thumbnails: Job error: TypeError: config.sizes is not iterable


zigmhount:
I'm having some issues with the aspect ratio of thumbnails, probably similar to #212. I replaced some pictures and tried to regenerate the thumbnails by clicking on the corresponding button in the advanced config, but I'm consistently getting `Job error: TypeError: config.sizes is not iterable`:
![image](https://user-images.githubusercontent.com/58641590/158671547-29da6cda-a83a-4155-8d82-68f19b94f5b0.png)

Running pigallery2 with the docker image nightly-debian-buster, with a different user than root which has read access to the pictures hosted on Nextcloud (user www-data, ID 33) - the issue is the same when running the container with the default root user :
```
version: '3'
services:
  pigallery2:
    image: bpatrik/pigallery2:nightly-debian-buster
    container_name: pigallery2
    environment:
      - NODE_ENV=debug
      - PORT=8080
    user: "33:33"    # www-data
    volumes:
      - "/mnt/raid1/pigallery2/config:/app/data/config" # Changed
      - "/mnt/raid1/pigallery2/db-data:/app/data/db"
      - "/mnt/raid1/clouddata/path/to/photos1:/app/data/images/photos1:ro"
      - "/mnt/raid1/pigallery2/tmp:/app/data/tmp" # Changed
    ports:
      - 3000:8080
    restart: always
```
**Server logs (docker)**
```
3/16/2022, 7:26:36 PM[VERBS] GET /api/admin/statistic 304 59ms
3/16/2022, 7:26:36 PM[VERBS] GET /api/user/list 304 10ms
3/16/2022, 7:26:36 PM[VERBS] GET /api/share/list 304 18ms
3/16/2022, 7:26:36 PM[VERBS] GET /api/admin/jobs/scheduled/progress 304 14ms
3/16/2022, 7:26:36 PM[VERBS] GET /api/admin/jobs/available 304 8ms
[Typeconfig] Loading defaults from file: undefined
[Typeconfig] Loading defaults, def prefix: default
[Typeconfig] no default cli found among these: {
	"expose": {
		"gc": true
	},
	"config": {
		"path": "/app/data/config/config.json"
	}
}
[Typeconfig] from env: {
	"Server": {
		"Database": {
			"dbFolder": "/app/data/db"
		},
		"Media": {
			"folder": "/app/data/images",
			"tempFolder": "/app/data/tmp"
		}
	}
}
[...]

PR #552, opened 2022-11-13
Update thumbnail.settings.component.ts

mrtnbr:
Fix for https://github.com/bpatrik/pigallery2/issues/534 and https://github.com/bpatrik/pigallery2/issues/458 as suggested by [@tilmanginzel](https://github.com/tilmanginzel) in #458.

I've been using this fix for a while and have not found any issues and/or side effects.

### DIFF
--- a/src/frontend/app/ui/settings/thumbnail/thumbnail.settings.component.ts
+++ b/src/frontend/app/ui/settings/thumbnail/thumbnail.settings.component.ts
@@ -50,7 +50,7 @@ export class ThumbnailSettingsComponent
   }
 
   get Config(): { sizes: number } {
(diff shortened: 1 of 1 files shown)


src/frontend/app/ui/settings/thumbnail/thumbnail.settings.component.ts
```diff
@@ -50,7 +50,7 @@ export class ThumbnailSettingsComponent
   }
 
   get Config(): { sizes: number } {
```
bpatrik, line 52: this.states.client.thumbnailSizes is an array (but here does uses the any type) The return value fo the function should be also `number[]`

PR #561, opened 2022-11-24
Also add heif

martadinata666:
Adding `.heif` as supported image type.

Thanks.


bpatrik:
adding `heic` was a mistake. I will revert that comment. 
We need to add a different libvips and update the metadataloader to work with it.

martadinata666:
ah i see, im gonna close this then. thanks for the info

PR #571, opened 2022-12-17
[Feature] add control for slideshow playback speed

### LINKED ISSUE #570: [Feature] Changing slideshow playback speed
I want to adjust the playback speed of a slideshow, for example to use this software for a DIY digital photo frame (my use-case).

When I open a lightbox, there should be a dropdown-control that allows me to choose a duration how long each image is shown, instead of just a fast-forward button.
(optional) This selection should ideally be stored and automatically applied, when I return to this page.

**Describe alternatives you've considered** (optional)
- Multiple buttons with different durations instead of a dropdown, but that get's crowded very fast or allows a lot less options
- Implementing playback speed with a setting in backend, but that would make every client have the same playback speed, be configured by an admin and not an user and not be changeable on the fly. (Also that would require much more changes I guess :laughing: )

I will open a PR, I'd be happy for a merge or feedback on what needs to happen for a merge or why you don't want this feature.

crypt-o-warrior:
Resolves #570

### DIFF
--- a/src/frontend/app/ui/gallery/lightbox/controls/controls.lightbox.gallery.component.html
+++ b/src/frontend/app/ui/gallery/lightbox/controls/controls.lightbox.gallery.component.html
@@ -198,6 +198,7 @@
             title="auto play"></span>
       <span class="oi btn-group highlight control-button"
             dropdown
+            (isOpenChange)="$event ? pause() : play()"
             [dropup]="true">
         <div id="button-duration"
              dropdownToggle
@@ -213,7 +214,7 @@
             aria-labelledby="button-duration">
           <li *ngFor="let duration of playBackDurations"
               role="menuitem"
-              (click)="play(duration)">
+              (click)="setPlayBackDuration(duration)">
             <a class="dropdown-item"
                [class]="selectedPlayBackDuration === duration ? 'active' : ''">
               {{duration}}s
--- a/src/frontend/app/ui/gallery/lightbox/controls/controls.lightbox.gallery.component.html
+++ b/src/frontend/app/ui/gallery/lightbox/controls/controls.lightbox.gallery.component.html
@@ -34,7 +34,7 @@
               aria-controls="dropdown-basic">
         <span class="oi oi-menu"></span>
       </button>
-      <ul id="dropdown-basic" *dropdownMenu class="dropdown-menu  dropdown-menu-right"
+      <ul id="dropdown-basic" *dropdownMenu class="dropdown-menu dropdown-menu-right"
           role="menu" aria-labelledby="button-basic">
         <li role="menuitem">
           <a *ngIf="activePhoto"
@@ -63,7 +63,7 @@
             </bSwitch>
           </div>
         </li>
-        <li role="menuitem ">
+        <li role="menuitem">
           <div class="dropdown-item d-flex justify-content-between">
             <span title="key: a" i18n-title i18n>Show faces</span>
             <bSwitch
@@ -83,7 +83,7 @@
         </li>
         <li
           *ngIf="activePhoto && activePhoto.gridMedia.isVideo()"
-          role="menuitem ">
+          role="menuitem">
           <div class="dropdown-item d-flex justify-content-between">
             <span title="key: l" i18n-title i18n>Play videos</span>
             <bSwitch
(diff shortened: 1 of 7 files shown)


src/frontend/app/ui/gallery/lightbox/controls/controls.lightbox.gallery.component.html
```diff
@@ -188,19 +188,41 @@
...
       <span
         class="oi oi-media-play highlight control-button"
         [ngClass]="playBackState == PlayBackStates.Play ? 'button-active':''"
         (click)="play()"
         title="auto play"></span>
-      <span class="oi oi-media-skip-forward highlight control-button"
-            [ngClass]="playBackState == PlayBackStates.FastForward ? 'button-active':''"
-            (click)="fastForward()"
-            title="fast auto play"></span>
+      <span class="oi btn-group highlight control-button"
```
bpatrik, line 200: oi only needed for the icon tag

src/frontend/app/ui/gallery/lightbox/controls/controls.lightbox.gallery.component.html
```diff
@@ -188,19 +188,41 @@
...
+            dropdown
+            (isOpenChange)="$event ? pause() : play()"
+            [dropup]="true">
+        <div id="button-duration"
+             dropdownToggle
+             data-bs-auto-close="outside"
+             aria-controls="dropdown-basic">
+          <span class="text-white pe-1">{{selectedPlayBackDuration + 's'}}</span>
+          <span class="oi-clock"></span>
```
bpatrik, line 209: so the oi is needed here

bpatrik:
looks good to me. thank you for your contribution

src/frontend/app/ui/gallery/lightbox/controls/controls.lightbox.gallery.component.html
bpatrik: never mind. Ill fix it.

bpatrik:
I found the design of the speed selector to be inconsistent with design of the app, so I moved to the submenu on the top.

Sorry about that. Please consult next time before sending a PR.

crypt-o-warrior:
Thanks for merging and fixing those things you highlighted! I had some trouble aligning the dropdown with those other buttons and 'io' added some weird position offset that affected height. Just got it to work like that, but I like the position you moved it to a lot, so doesn't really matter anymore I guess.

I'm happy to contribute, this was actually my first time ever :partying_face: What do you mean by 'consult first'? I thought opening a PR and an issue is just that.

I do have a few improvements that I'd like to see. You removed some logic by refactoring the dropdown at a new location. Should I open separate issues for that or would you like to discuss those here and I open issues/PRs afterwards?

What I'd like to re-add:
1. Set cookie for playback speed. Especially since it's now in the submenu this would be very handy, because it's a little hidden now and would be great as a set-and-forget kind of setting. At least per client and if possible I'd set expires to 10 years or something, but I could not get that to work, at maximum the cookie was created with 7 days validity. 

This feature was there already, but since you are using two-way-data binding now, my function that also set a cookie isn't called anymore. I have no idea why changing playback speed works while having an active slideshow running. The logic you added is also in that function and I just can't see it called anywhere and if this works, the cookie should be set aswell. Maybe you see something I don't.
 
2. Selecting a playback speed should start the slideshow (arguable, but I personally think if someone is setting a playback speed, they most likely want to start it next)

Thank you for your work on this! I'm more than happy to fix those things myself and create a PR after we decided what to do

bpatrik:
Hi

By consulting I meant it is better to have some back and forth mail before someone puts effort into implementing something. (I hope I did not sound harsh, that was no my intention).
The May reason behind it is that it's basically only who has the whole code base in my head.
Also I'm relatively picky on the design. Like I won't submit half solutions:)

About the issues:

1. Sorry I did not see that the change broke the cokie settings. I will need to double check if directory ordering uses local storage. If yes this feature should also do so.

2. I'm not 100% that settings is at the right place, but could figure out any better . I might move it back to the play button but just something super minimalistic. Or make the play button behave like the FB like button where longpress/hover opens the other like options.
I do not see this feature to be used often. As you said it should be set and forget. I think the default value should be also set in the config

PR #578, opened 2023-01-08
Dutch translation

mrtnbr:
I also found a few fields that could not be translated ie. are not in the translation file, mainly stuff in the Search dialogue. I will open an issue for those.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -65,6 +65,10 @@
           "pl": {
             "baseHref": "",
             "translation": "src/frontend/translate/messages.pl.xlf"
+          },
+          "nl": {
+            "baseHref": "",
+            "translation": "src/frontend/translate/messages.nl.xlf"
           }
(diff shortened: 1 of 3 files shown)


mrtnbr:
I'm not sure how to exclude the change to the Dockerfile I made for my own testing, sorry I'm a bit of a Git newbie.

bpatrik:
> I'm not sure how to exclude the change to the Dockerfile I made for my own testing, sorry I'm a bit of a Git newbie.

Its not part of this PR, so looks good to me.

bpatrik:
> I also found a few fields that could not be translated ie. are not in the translation file, mainly stuff in the Search dialogue. I will open an issue for those.

Thanks. there are some search related text that cannot be translated at the moment. But let me know if you find something that is not translated.

PR #583, opened 2023-01-09
A few brazilian portuguese translations - To Be Continued

marcossatoshi:
I've started the translation for a new language but didn't know if it was a problem using pt-br instead of ptbr. Since there's also pt-pt and had no other examples of dualities I went for ptbr without the -

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -69,6 +69,10 @@
           "nl": {
             "baseHref": "",
             "translation": "src/frontend/translate/messages.nl.xlf"
+          },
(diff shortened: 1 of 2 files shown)


bpatrik:
Here should be the list of possible locales: https://angular.io/guide/i18n-common-locale-id#refer-to-locales-by-id

Based on that I think it should be with `-`, but needs to be tested. 


I will merge this and take a look

bpatrik:
It works if I add `-`: 
kép

PR #597, opened 2023-01-25
The spaces for ffmpeg custom options are strict.

lijunle:
After some debugging, the custom options for fluent-ffmpeg are strict with spaces. We cannot have prefix or suffix spaces. The hint is confusing because it shows a space ` ` after semicolon `;`.


bpatrik:
Thank you!

Issue #177, opened 2020-09-02
custom sorting method for a single directory


kolydart:
First of all, thanks for the great package. It is really helpful and has given to me and my family many happy moments.

**Feature description**
I'm trying to figure out a way to apply custom sorting method on root directory. For me, the most obvious way to browse a picture gallery is having the latest folders on top in the root directory and the files/folder inside each folder sorted by  ascending date. 

**Proposed solution**
A specific directory should have a way to override the general settings for sorting methods. I think the best (simplest) way would be by having a hidden file (ex. `.order_descending_name`).

**More info on my solution**
I have all my pictures in folders with prepending date, (ex. `2010.09.02 John's wedding`). Sorting the root directory by descending name is the absolutely best approach for the gallery root; recent events are on top. 
But, if I set that method to settings,  photos inside each folder are in reverse order, having the latest on top, which is not helpful.

Thanks

bpatrik:
Hi,

I'm not sure if I fully understand your request.
Would you like to have:
1) Sort subfolders descending and photos ascending in the same folder?
2) Sort one folder descending and photos ascending in a separate folder?

For 1) I'm not sire if I would like to add this complexity to  my app. You can convince me if an other photo does it in a elegant way.
For 2) That should be mostly possible now too. You can have a default sorting setting and override that in the individual folders.

kolydart:
Case 2 was what I was talking about. Sorting by "ascending date" as default (in each folder under root) and "descending name" in a single folder as override (root only).
I was not aware it can be done.

So, how can I **permanently** override a specific folder's sorting? 
If I change a folder's  sorting method using the upper right dropdown button (button#button-alignment) and visit the same folder in another browser, the sorting order is back to the default.

thank you!

bpatrik:
yes, the sorting is stored in the browser's cache, so in a different browser is gets back to the default value.

Unfortunately there is no permanent way of doing this currently.

kolydart:
So, would there be any chance of implementing it?

bpatrik:
Alright. I had some time and added this functionality. :) Here you can see how it works: https://github.com/bpatrik/pigallery2/blob/73a8c4202a9d4fd7746e080a339c0fa644d4b32f/src/common/PG2ConfMap.ts#L10-L14

Basically, you add an empty file like `.order_descending_name.pg2conf` to a folder and that overrides the default sorting for that folder.

kolydart:
that was fantastic!
thanks!

[...]

PR #620, opened 2023-02-24
document issue #177 about overriding default sort order

kagahd:
The feature to override the default sort order was not yet documented, so I added it.


bpatrik:
Thank you. The documentation part of the app is not that clean :/
I could not figure out for a long time. Lately I try to make the app self documented and keep all the options in the app. (i.e.: documentations like this should show app in the app eventually). Like I will add a popup in the settings around sorting that would show the same text.

Issue #630, opened 2023-03-09
Password protected sharing broken, errors "Unknown sharing key"


jdavidpeter:
**Describe the bug**

When running PiGallery, password protected share creates links those are served by an "Unknown sharing key." error page.
An example link in my case: https://redacted.duckdns.org/shareLogin?sk=a9322371

When disabling password protection, the sharing feature works just fine.

When password protection is disabled, a temp user is used to access the shared links. If I reenable password protection while browsing as a temp user, that user session can still access all the password protected shares without prompt for a password.
If I log out from the temp user during browsing a password protected share, I'm redirected to a login page where only a password is requested for that share. Typing the right password allows me to access the share as expected, however, if I use the generated share link, I'm again greeted by "Unknown sharing key."

**Photo/video (optional) that causes the bug**

Bug is not connected to any particular media.

**Screenshots (optional)**

![image](https://user-images.githubusercontent.com/14619409/224014938-81e1c77a-f7d8-434a-9344-eaf7d2618bf7.png)

**Server logs (optional)**

```
redacted@redacted:~/pigallery$ docker-compose logs --tail=8
Attaching to pigallery2, nginx
nginx         | 81.183.73.163 - - [09/Mar/2023:11:40:40 +0000] "POST /pgapi/user/logout HTTP/1.1" 200 59 "https://redacted.duckdns.org/shareLogin?sk=a9322371" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:109.0) Gecko/20100101 Firefox/110.0"
nginx         | 81.183.73.163 - - [09/Mar/2023:11:42:19 +0000] "GET /shareLogin?sk=a9322371 HTTP/1.1" 200 2673 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:109.0) Gecko/20100101 Firefox/110.0"
nginx         | 81.183.73.163 - - [09/Mar/2023:11:42:19 +0000] "GET /pgapi/share/a9322371 HTTP/1.1" 401 130 "https://redacted.duckdns.org/shareLogin?sk=a9322371" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:109.0) Gecko/20100101 Firefox/110.0"
[...]

PR #633, opened 2023-03-12
Attempt to fix not accessible password protected shares

jdavidpeter:
https://github.com/bpatrik/pigallery2/issues/630

### DIFF
--- a/src/frontend/app/ui/gallery/share.service.ts
+++ b/src/frontend/app/ui/gallery/share.service.ts
@@ -131,8 +131,11 @@ export class ShareService {
       );
       this.sharingSubject.next(sharing);
     } catch (e) {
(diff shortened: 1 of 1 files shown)


jdavidpeter:
This solution fails to provide the correct Unknown Sharing Key error page even if I try to access a non-existing share, so likely it's not good enough.

bpatrik:
The sharing login is rather complex part of the app. I recommend doing a TDD approach and writing clear scenarios as test cases. They should go somewhere here: https://github.com/bpatrik/pigallery2/blob/master/test/backend/integration/routers/SharingRouter.ts

bpatrik:
this should be fixed by now.

jdavidpeter:
Thank you!

Issue #682, opened 2023-07-27
Use of UTC for video files


mcwieger:
Video files do not appear to use UTC, but seem to display the GMT timezone value.

I set the following (Quicktime) tags to make sure all possible dates used for videos are correct:

```
[System]        FileModifyDate                  : 2023:04:21 18:53:30+02:00
[System]        FileAccessDate                  : 2023:07:27 19:58:17+02:00
[System]        FileCreateDate                  : 2023:04:21 18:53:30+02:00
[QuickTime]     CreateDate                      : 2023:04:21 18:53:30+02:00
[QuickTime]     ModifyDate                      : 2023:04:21 18:53:30+02:00
[ItemList]      ContentCreateDate               : 2023:04:21 18:53:30+02:00
[Track1]        TrackCreateDate                 : 2023:04:21 18:53:30+02:00
[Track1]        TrackModifyDate                 : 2023:04:21 18:53:30+02:00
[Track1]        MediaCreateDate                 : 2023:04:21 18:53:30+02:00
[Track1]        MediaModifyDate                 : 2023:04:21 18:53:30+02:00
[Track2]        TrackCreateDate                 : 2023:04:21 18:53:30+02:00
[Track2]        TrackModifyDate                 : 2023:04:21 18:53:30+02:00
[Track2]        MediaCreateDate                 : 2023:04:21 18:53:30+02:00
[Track2]        MediaModifyDate                 : 2023:04:21 18:53:30+02:00
```
All times are set to 18:53:30 with my timezone. However, date in PiGallery2 shows 16:53:30. Can UTC be implemented for video files as it was done for pictures in https://github.com/bpatrik/pigallery2/pull/480 and https://github.com/bpatrik/pigallery2/issues/469?

Environment
 - OS: Synology DSM6.2
 - Browser: Chrome

Used app version
- Docker Nightly Alpine v1.9.6-nightly

bpatrik:
I think the problem lies with the metadata parsing part: The `creationDate` is not set correctly most likely here: https://github.com/bpatrik/pigallery2/blob/3e6bb7bff719f2c768c8ddf939cc63089a8f7776/src/backend/model/threading/MetadataLoader.ts#L20-L93

mcwieger:
This is what Phil Harvey says: https://exiftool.org/forum/index.php?topic=15052.0

The Quicktime creation date indeed is not using the right timezone. If that field can be made to use UTC, the time will be correct.

[...]

PR #710, opened 2023-09-03
Matroska container header meta data parsing

GenericGuy:
From what I've gathere online, Matroska (mkv) files don't provide the same interface as MP4 files for reading metadata. Duration and creation date are not stored in the stream section, but in the container section.

Mainly 3 additions:
- Duration is read from the format section only if it's not set already (haven't found too much information on this, but based on my personal files the stream metadata seems to be more accurate for the duration)
- Bit rate from the format section overwrites previously found bit rates (since the stream bit rate really only accounts for video, it makes more sense to use the average overall bit rate)
- Creation date also overwrites previous creation dates (found no file in my collection where these two were not the same).

I've also added a testcase for mkv metadata parsing and updated the bit rates for the other testcases.

tldr: before this PR any mkv files in my library would use the filesystem modification time, now the embedded creation time is used.


bpatrik:
Thank you for the clean PR.

This might also help with #682 (need to check)

Issue #711, opened 2023-09-03
Improve blog experience: add support for per day markdown


bpatrik:
This a tracking bug for a feature for adding support to the blog feature to be able to create text that is attached to a date.
It only shows if you group by date. Otherwise those date related markdown will show at the main markdown.

Sample:
![kép](https://github.com/bpatrik/pigallery2/assets/7512182/d88cd29b-110e-459d-96dc-8fc01aac47ca)

This effort is part of #587

PR #712, opened 2023-09-03
Inline or groupped blog #711

bpatrik:
You can tag section in the `*.md` files  with `<!-- @pg-date <ISO_DATE> -->`, like: `<!-- @pg-date 2015-06-12 -->` to attach them to a date. 
Then, if you group by date, they will show up at the assigned day.

That part of the markdown will be removed from the main markdown at the top and shown only at that day.

### DIFF
--- a/demo/images/index.md
+++ b/demo/images/index.md
@@ -83,4 +83,11 @@ Start numbering with offset:
 57. foo
 1. bar
 
+<!-- @pg-date 2015-06-12 -->
+## Day 1
+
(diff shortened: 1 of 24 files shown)


knalltuete86:
Dear @bpatrik 
I just saw this Feature at the Demo instance and updated my instance from source to latest (1.9.6-edge).

Conclusion: Great feature!

Question: Is there any way to extend the Markdown within the Gallery by default? Currently I always have to click once for each section to show it ...

Regards

bpatrik:
[3825c41](https://github.com/bpatrik/pigallery2/commit/3825c41f13ba0dc73e8635ce72f65eaa508b155a) should add a config to make it open by default.

cheers.

PR #761, opened 2023-11-14
Add XMP sidecar support in backend for video files

grahamalderson:
This pull request adds xmp sidecar support using exifr in order to gain access to keywords metadata for video files. It can be extended to collect additional data for videos or photos and is a first step towards reducing the number of exif readers per #277 .

This is a first step towards #156 
Exifr was chosen per #277 
Work accelerated by reviewing exiftool code from @mcdamo on #294

Known limitation of this pull request:
-Keywords for videos do not show up in the info panel but tagged videos are returned when you search for the keyword.

It's unclear to me why the tags do not show up in the info panel on the videos. **If you search for keyword:test or keyword:rabbit you will get back the bunny.mp4 video**, suggesting they are correctly recorded in the database. 

I attempted modification of frontend files to identify video vs photo and load metadata separately for them on the info panel believing that to be the issue, but the problem was deeper. When logged to console in the browser, metadata.keywords was showing as undefined. Given this, it's not clear my modifications to the frontend were even necessary, and I have not included them. I would be happy to look at it further with some pointers on why metadata.keywords is returning undefined on the frontend for videos despite them returning in keyword search results.

### DIFF
--- a/src/backend/model/fileaccess/MetadataLoader.ts
+++ b/src/backend/model/fileaccess/MetadataLoader.ts
@@ -12,6 +12,8 @@ import {IptcParser} from 'ts-node-iptc';
 import {FFmpegFactory} from '../FFmpegFactory';
 import {FfprobeData} from 'fluent-ffmpeg';
 import {Utils} from '../../../common/Utils';
+import * as exifr from 'exifr';
+import * as path from 'path';
 
 const LOG_TAG = '[MetadataLoader]';
 const ffmpeg = FFmpegFactory.get();
@@ -30,6 +32,29 @@ export class MetadataLoader {
         fileSize: 0,
         fps: 0,
       };
+      
+      try {
+        // search for sidecar and merge metadata
+        const fullPathWithoutExt = path.parse(fullPath).name;
+        const sidecarPaths = [
+          fullPath + '.xmp',
+          fullPath + '.XMP',
+          fullPathWithoutExt + '.xmp',
+          fullPathWithoutExt + '.XMP',
+        ];
+
+        for (const sidecarPath of sidecarPaths) {
+          if (fs.existsSync(sidecarPath)) {
+            const sidecarData = exifr.sidecar(sidecarPath);
+            sidecarData.then((response) => {
+              metadata.keywords = [(response as any).dc.subject].flat();
+            });
+          }
+        }
+      } catch (err) {
+        // ignoring errors
+      }
+
       try {
         const stat = fs.statSync(fullPath);
         metadata.fileSize = stat.size;
--- a/src/backend/model/fileaccess/MetadataLoader.ts
+++ b/src/backend/model/fileaccess/MetadataLoader.ts
@@ -32,7 +32,7 @@ export class MetadataLoader {
         fileSize: 0,
         fps: 0,
       };
-      
+
       try {
         // search for sidecar and merge metadata
         const fullPathWithoutExt = path.parse(fullPath).name;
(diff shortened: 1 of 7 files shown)


bpatrik:
Hi,
first thank you for the contribution. This is a long outstanding issue. I wanted to refactor the whole metadata loader class, but never got there.

Unfortunately exifr is not maintained anymore https://github.com/MikeKovarik/exifr (last commit 2 years ago)

I'm still fine with this contribution at this point as it at least move forward the issue.

You only implemented sidecars for videos. They wont be loaded for jpgs. Can you also add that?

Video keywords are not showing up as they ar filtered out at server side:
https://github.com/bpatrik/pigallery2/blob/237e61c35083ddc73c2ef47ab1fb78c4e3c593ac/src/common/entities/ConentWrapper.ts#L244

grahamalderson:
I missed that exifr was no longer maintained, that's too bad. I had been planning to slowly work through refactoring metaloader loader and implement exifr for everything, but I guess that's not a good idea at this point I guess so I won't proceed.

Thanks for the pointer on the contentwrapper, I think I can get that done today or tomorrow.

I'm happy to give sidecars for photos a try in this PR as well. I had been planning to do that in the future. It wasn't clear to me if that was something that was widely used since XMP can be contained in the photo file, so I didn't prioritize it.

If we see different values between the metadata in the photo file and the metadata in the sidecar, which should we prefer? It's less of an issue while I'm just working on keywords right now since we can just concatenate them, but for the next step as sidecar support is extended, what if we see conflicting values?

I'll submit a future PR to extend the sidecar support beyond keywords.

Thanks for building this and the quick response!

grahamalderson:
@bpatrik Outstanding issues resolved.
-video keywords now show up on frontend
-sidecar support extended to photos same as in videos

This PR is still limited to support for keywords in sidecars, but can be extended further in a future PR

bpatrik:
Thank you very much!

>  It's less of an issue while I'm just working on keywords right now since we can just concatenate them, but for the next step as sidecar support is extended, what if we see conflicting values?

Sidecar should override the photo. I think that is expected if you have an explicit sidecar next to your photo that it is there for a reason and should be used as the authoritative source of truth. If you want to make it even better, you can create a settings for `override`, `skip`,`merge` to support all possible scenarios. But I do not find that necessary for now.

src/backend/model/fileaccess/MetadataLoader.ts
```diff
@@ -30,6 +32,29 @@ export class MetadataLoader {
...
+          fullPath + '.xmp',
+          fullPath + '.XMP',
+          fullPathWithoutExt + '.xmp',
+          fullPathWithoutExt + '.XMP',
+        ];
+
+        for (const sidecarPath of sidecarPaths) {
+          if (fs.existsSync(sidecarPath)) {
+            const sidecarData = exifr.sidecar(sidecarPath);
+            sidecarData.then((response) => {
```
bpatrik, line 49: It occurred to me that this is an async call that you are not waiting for.

So there might be some race condition when the sidecar won't be properly loaded.

src/backend/model/fileaccess/MetadataLoader.ts
```diff
@@ -155,6 +180,28 @@ export class MetadataLoader {
...
+                fullPath + '.xmp',
+                fullPath + '.XMP',
+                fullPathWithoutExt + '.xmp',
+                fullPathWithoutExt + '.XMP',
+              ];
+      
+              for (const sidecarPath of sidecarPaths) {
+                if (fs.existsSync(sidecarPath)) {
+                  const sidecarData = exifr.sidecar(sidecarPath);
+                  sidecarData.then((response) => {
```
bpatrik, line 196: Same applies to here too. Async call that is not waited for.

grahamalderson: I learned it was asynchronous the hard way when getting unreliable results in testing due to the race condition. I thought I resolved that by wrapping it in the .then clause which causes a wait for the promise to be fulfilled, my testing seemed reliable afterwards. Happy to resolve whatever I have overlooked, just unclear on what is still missing as I'm pretty new to js.

bpatrik: sidecarData has the value of Promise<some type>.

Promises have values sometimes in the future. You can use then to subscribe to the event. The problem that the execution will continue. The main function can also return. And eventually in the future the then function will run.

I recommend you trying to refactor the main function to be async. So you can use await on a promise. 
I could not find a good blog now, but something like this might help: https://blog.logrocket.com/async-await-in-typescript/

grahamalderson: Thanks for taking the time to provide the pointer. That was helpful, as was a search for ".then vs await". 

I think I understand the problem you're highlighting. By using .then I fixed the immediate race condition for setting the metadata.keywords before the sidecar load completed, but the race condition for the overall return on video metadata before it sets keywords is still present since .then allows processing to continue while waiting on this. I'm not seeing the effect in testing only because so far the subsequent ffmpeg processing is never faster than the sidecar load. We can't rely on that, so await is a better approach. 

I'll find some time soon to refactor appropriately.

grahamalderson: Pull #792  for resolution of refactoring

PR #764, opened 2023-11-15
add description for album reset

kagahd:
I didn't know that an `album reset` would delete all saved searches (albums), so I lost all my albums with a single click.
To prevent other PiGallery2 users from that mistake, I just added a description to `album reset`.

### DIFF
--- a/src/common/config/public/ClientConfig.ts
+++ b/src/common/config/public/ClientConfig.ts
@@ -1457,6 +1457,7 @@ export class ClientConfig {
         hideProgress: true
       }]
     } as TAGS,
+    description: $localize`An album reset deletes all albums (all saved searches).`
   })
   Album: ClientAlbumConfig = new ClientAlbumConfig();
 


bpatrik:
I think we should just rather rename the job to "Delete albums" to make it obvious.
Also there is already a warning if you add the action in the jobs tab:

kép

kagahd:
> Also there is already a warning if you add the action in the jobs tab:

As you showed in your screenshot, it seems that the **job** has an according description. 
However, it's too easy to just click on the red `Album reset` button as shown in the following screenshot. As you can see in that screenshot, there is no description what will happen on pushing neither of these buttons. A description is therefore very welcome.
albumReset

kagahd:
By the way, what does `Faces` do? There is neither a description nor a `Faces` job that one could add in hope to get a description there.
faces

kagahd:
It seems that the content of the `Faces` panel is only shown in `Under the hood` mode as the following screenshot demonstrates. So please disregard my previous comment or copy it to a dedicated issue.
faces_underTheHood

bpatrik:
> > Also there is already a warning if you add the action in the jobs tab:
> 
> As you showed in your screenshot, it seems that the **job** has an according description. However, it's too easy to just click on the red `Album reset` button as shown in the following screenshot. As you can see in that screenshot, there is no description what will happen on pushing neither of these buttons. A description is therefore very welcome. albumReset

Yeah I agree that there is room for improvement.

But instead adding more text to the panel, I would rather:
1. rename the job (that button starts a nameless job. they are the same) to be more obvious. 
2. When clicking on a red /" destructive" button, a a popup should ask "are you sure? this will: <job description>".

These changes would be client side only. If you willing to give it a try, I would appropriate it. Otherwise, please open a bug and I will add it before the next release.

kagahd:
>  a popup should ask "are you sure? this will: <*job description>".

@bpatrik, do you have such a popup already elsewhere in place, so I can copy it to keep the popup consisting with existing ones? If not, do you have any specifications on how it should look and work?

bpatrik:
The app uses ngx-bootstrap, you can use them to create a popup modal https://valor-software.com/ngx-bootstrap/#/components/modals?tab=overview

bpatrik:
albums add modals: https://github.com/bpatrik/pigallery2/blob/d38830f8be333466ebf44e79e5f1c8c1382aca20/src/frontend/app/ui/albums/albums.component.html#L31

Also see the .ts file

bpatrik:
[dccd9c3](https://github.com/bpatrik/pigallery2/commit/dccd9c30e193649cae471b5d6e0f29509956c71b) should fix this.

PR #765, opened 2023-11-18
Some German translations

marcusfey:
Version 2.0.0 brings a few very nice features, I like pigallery2 more than ever!

Hopefully my translations can help a bit. Yes, not too much, but now some non-admin-texts are translated again. I might add more, whenever I find the time

I used the word "Anwendung" (en: application) where "app" was used in English. Although "app" gets into German more and more it's still not as clear as in English, especially with web apps. I can change that, if required, or just do it yourself.


bpatrik:
Thank you very much!

I don't have a strong opinion about the naming conventions. You can also use 'Seite' i guess, but I'm not native.

PR #771, opened 2023-11-22
Some German translations, pt.3 (of ?)

marcusfey:
I missed some texts visible only to normal users in #765. Surely Germanists would kill me for some translations...

Additionally now some more texts under settings. I'm unsure how much time I should invest here because admins would usually cope with English.


bpatrik:
I would not stress too much on the setting translation. The one who sets it up, needs to minimally understand english to read the docs here. Also settings change more often then normal UI.

bpatrik:
Also, thank you for the contribution :)

Issue #776, opened 2023-11-24
Feature request: store filters in local storage


josiah-roberts:
**Is your feature request related to a problem? Please describe.**
I'd like for PiGallery to preserve the filters I previously set when I come back to the page. The particular reason here being to maintain a filter that drops photos with no date info, since they get sorted to the top (an unrelated issue). I can solve this by having the filters persisted in local storage.

**Describe the solution you'd like**
Whenever filters are updated, persist those changes to local storage. When the page is loaded, pull that filter from local storage if present.

**Describe alternatives you've considered**

- User-level server side persistence of filters
  This is super heavy-handed and complex
- Use url-query parameters
  This isn't nuts, but it doesn't fully do what I want (come back to where I was when I navigate to http://mygallery), and it's more heavy-handed and represents a significantly larger change

bpatrik:
I was thinking adding the 2nd alternative to the app. So when you share a folder it would possible to share the actual view that you are looking at.

Alternative 3 would be to use `pg2conf` files: src/common/PG2ConfMap.ts
User creates a *.pgconf file in the directory. That file auto goes to the client side and the client can read it and do something based in that.

------------------------------------

I'm not 100% I want this feature in the app, but I'm not super against it. I see that you started working on some implementation. Can you please put this behind some flag (config) and make it disabled by default.

I'm doing something similar with sorting, if you want some pointers: https://github.com/bpatrik/pigallery2/blob/fc75be1b25f012b721c44238cac38d5f253cd2ac/src/frontend/app/ui/gallery/navigator/sorting.service.ts#L60-L67
There is GalleryCacheService that interacts with local storage.

josiah-roberts:
> I'm not 100% I want this feature in the app, but I'm not super against it. I see that you started working on some implementation. Can you please put this behind some flag (config) and make it disabled by default.

Yeah, if/when I actually get to any behavior changes I'll config-flag it.

> I'm doing something similar with sorting, if you want some pointers:
> https://github.com/bpatrik/pigallery2/blob/fc75be1b25f012b721c44238cac38d5f253cd2ac/src/frontend/app/ui/gallery/navigator/sorting.service.ts#L60-L67

Yes, behaviorally this is essentially what I was imagining for filters.

bpatrik:
There is one more thing that worth considering when implementing this feature. I plan to add the following features sometime in the future:
1. filter photos on the server side (like hide some photos for certain users) This would also unblock sharing search results
[...]

PR #778, opened 2023-11-24
Disentangle filter types + library counts + selections

josiah-roberts:
Based on #780

## Why

Currently, the active filter state intermingles:
- Filter type implementation: `mapFn`, `name`, etc.
- Library data: count of items with various filter values
- UI data: selected filter types and options

Because of this, even starting on #776 isn't really feasible.

## What

This PR separates out:
- Filter type implementation: `FilterService.filters` + `FilterService.filtersMap`
- Library data: `FilterService.filterState.filterValueCounts`
- UI data: `FilterServie.filterState.selectedFilters`

## What else

~~This also fixes an odd issue where the leftward filters impact the counts on filters to their right, but not vice-verse. Now filters all report all values in the date-filtered photo set. **If this is intended**, I can update it to be consistent instead of left-to-right.~~ This _was_ intended - I just misunderstood the desired behavior.

- [ ] Revert change to intended left-to-right filter behavior

Additionally, I've addressed a number of inconsistencies and type issues - removing casts and making `mapFn` always return `string[] | undefined`.

### DIFF
--- a/src/frontend/app/ui/gallery/filter/filter.gallery.component.ts
+++ b/src/frontend/app/ui/gallery/filter/filter.gallery.component.ts
@@ -1,6 +1,6 @@
 import {Component, OnDestroy, OnInit} from '@angular/core';
 import {RouterLink} from '@angular/router';
-import {FilterOption, FilterService, SelectedFilter} from './filter.service';
+import { Filter, FilterService, FilterState, FilterType, SelectedFilter} from './filter.service';
 
 @Component({
   selector: 'app-gallery-filter',
@@ -40,20 +40,27 @@ export class GalleryFilterComponent implements OnInit, OnDestroy {
     );
   }
 
-  get ActiveFilters(): {
-    filtersVisible: boolean;
-    areFiltersActive: boolean;
-    dateFilter: {
-      minDate: number;
-      maxDate: number;
-      minFilter: number;
-      maxFilter: number;
-    };
-    selectedFilters: SelectedFilter[];
-  } {
-    return this.filterService.activeFilters.value;
+  get ActiveFilters(): FilterState {
+    return this.filterService.filterState.value;
   }
 
+  filterChanged(index: number, newFilter: Filter) {
+    this.filterService.filterState.value.selectedFilters[index] = {
+      type: newFilter.key,
+      options: {}
+    }
+    this.filterService.onFilterChange();
+  }
+
+  toggleFilterValue(filter: SelectedFilter, value: string) {
+    filter.options[value] = !filter.options[value];
+    this.filterService.onFilterChange();
+  }
+
+  options(filter: SelectedFilter) {
+    const entries = Object.entries(filter.options).map(([k, v]) => ({selected: v, name: k}))
+    return entries.sort((a, b) => this.filterService.filterState.value.filterValueCounts[filter.type][b.name] - this.filterService.filterState.value.filterValueCounts[filter.type][a.name])
+  }
 
   ngOnDestroy(): void {
     setTimeout(() => this.filterService.setShowingFilters(false));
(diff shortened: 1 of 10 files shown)


src/frontend/app/ui/gallery/filter/filter.service.ts
```diff
@@ -1,98 +1,119 @@
...
+    mapFn: (m: PhotoDTO) => [m.metadata.positionData?.state],
+    renderType: FilterRenderType.enum,
+  },
+  country: {
+    key: 'country',
+    name: $localize`Country`,
+    mapFn: (m: PhotoDTO) => [m.metadata.positionData?.country],
+    renderType: FilterRenderType.enum,
+  },
+} as const;
```
josiah-roberts, line 85: I've pulled this out of the service class itself so that I can derive `type` declarations from it.

src/frontend/app/ui/gallery/filter/filter.service.ts
```diff
@@ -275,66 +300,56 @@ export class FilterService {
...
+                    );
+                  },
+                  false
+                );
+                if (!filteredOut) {
+                  yield item;
+                }
               }
+              return filterValueCounts;
+            }
```
josiah-roberts, line 337: This filters in a single pass over `c.media`. I don't love encoding some of the filter logic here though - e.g. "if there are no keywords, leave it in." I'd love to factor this out into the individual filters themselves.

bpatrik:
> This also fixes an odd issue where the leftward filters impact the counts on filters to their right, but not vice-verse. 

This is work as intended. The filters bar is actually a copy from lightroom and tries to mimic the same behavior. 
You can think of the filters as pipes, going from left to right.

bpatrik:
I find it hard to review this PR.

Can you let me know if this change does any:
1. Behavioral change
2. Any UI change?

Do I understand correctly this is only a refactor and should not have any changes?

Can you help the review with the following:
1. Write a test where its clear that this change only does a refactor? (maybe send in a separate PR so that we can see it passes with the current code?)
2. Or at least send this PR in smaller chunks so that is easier to parse


If there is a behavioral and or UI change. Please elaborate on them also include a screenshot that clearly shows the difference.

Sorry but after work I find it hard to gather the brain capacity to review complex PRs.

src/frontend/app/ui/gallery/filter/filter.gallery.component.ts
```diff
@@ -40,20 +40,22 @@ export class GalleryFilterComponent implements OnInit, OnDestroy {
...
-    dateFilter: {
-      minDate: number;
-      maxDate: number;
-      minFilter: number;
-      maxFilter: number;
-    };
-    selectedFilters: SelectedFilter[];
-  } {
-    return this.filterService.activeFilters.value;
+  get ActiveFilters(): FilterState {
```
bpatrik, line 43: Can you also rename the getter function to follow the property name?

josiah-roberts:
> Do I understand correctly this is only a refactor and should not have any changes?

Yes, that's correct with the exception of the left-to-right filter behavior, which I didn't understand was intended. I'll revert that change, and there shouldn't be any other behavior changes - it's purely intended to split apart filter behavior, selection data, and library state data.

> Can you help the review with the following:
> 
> 1. Write a test where its clear that this change only does a refactor? (maybe send in a separate PR so that we can see it passes with the current code?)
> 2. Or at least send this PR in smaller chunks so that is easier to parse

Happy to put up a test PR - do you have a preference between an E2E test vs. a unit test around the filter service?

> Sorry but after work I find it hard to gather the brain capacity to review complex PRs.

No worries - thanks for maintaining the project!

src/frontend/app/ui/gallery/filter/filter.gallery.component.ts
josiah-roberts: 64a4fc787f06c8c530e3d9eb866ab45868dd62a0

bpatrik:
> Happy to put up a test PR - do you have a preference between an E2E test vs. a unit test around the filter service?

Thank you. I fine with any of them.

bpatrik:
I also made some change in this space. See latest commit. I expect it to be an easy merge

bpatrik:
I will reject this PR in the following days due to lack of activity.

PR #795, opened 2023-12-04
Refactor sidecar loading

grahamalderson:
Addresses concerns in #792 except does not yet include merge function, scoped to future PR.

This fixes some edge case errors in sidecar loading that caused indexing to fail and application to crash. It also removes use of any in favor of defined types, and finishes async/await refactor.


bpatrik:
Thanks!

Issue #760, opened 2023-11-13
Showing both short and long captions in the sidebar


nk9:
### Is your feature request related to a problem? Please describe.
I have used digiKam to add both titles and captions to a bunch of photos. The digiKam UI looks like this:

<img width="384" alt="image" src="https://github.com/bpatrik/pigallery2/assets/3646730/0caeb63b-0d04-423c-b6fc-91070ba60cc7">

And according to https://jimpl.com/, it produces EXIF output like this:

Field | Value
-- | --
Title | Baby Olga
UserComment | Age 6 months
Caption | Baby Olga
Caption-Abstract | Age 6 months

Initially, neither of these fields were visible in pG2. Then I turned on "Caption first naming", and now I can see the "UserComment" or "Caption-Abstract" string as an overlay in the UI:

<img width="769" alt="image" src="https://github.com/bpatrik/pigallery2/assets/3646730/03c8e270-634e-4365-b534-9253a49d2e03">

But the "Title"/"Caption" is still missing. And to be honest, the "title" is what I'd probably expect to see as an overlay, with the longer caption visible in the metadata sidebar if a user opens it. The [IPTC spec says](https://iptc.org/std/photometadata/specification/mapping/iptc-pmd-newsmlg2.html) that, despite the misleading suffix "-Abstract," this is the "Description" field. So it's where I'd expect to see longer text associated with the image, vs the short, pithy "Title."

### Describe the solution you'd like
I would expect both the "title" and the "caption" (as named by digiKam, whichever non-duplicative fields these map to) to be visible in the metadata sidebar. This seems pretty obvious to me, to the point that I would argue that the current behavior could be considered broken. Even if one of the two caption fields is available in that potentially hard-to-read and definitely-not-selectable overlay, I'd still expect both to be visible and selectable in the sidebar. (If a photo doesn't have any of these fields, then I wouldn't expect them to show in the sidebar.)

[...]

PR #798, opened 2023-12-06
[#760] Add image titles and captions to the info panel.

nk9:


bpatrik:
Looks reasonable. Thank you!

PR #816, opened 2024-01-20
Issue 814 : Name the zip file after the folder name

fred1701d:
The name of the zip when downloading a folder should be named after the folder.
For root, default will be gallery.zip

### DIFF
--- a/src/backend/middlewares/GalleryMWs.ts
+++ b/src/backend/middlewares/GalleryMWs.ts
@@ -84,6 +84,7 @@ export class GalleryMWs {
     res: Response,
     next: NextFunction
   ): Promise<void> {
+    var contentDisposition = require('content-disposition')
     if (Config.Gallery.NavBar.enableDownloadZip === false) {
       return next();
     }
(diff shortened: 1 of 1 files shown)


src/backend/middlewares/GalleryMWs.ts
```diff
@@ -84,6 +84,7 @@ export class GalleryMWs {
     res: Response,
     next: NextFunction
   ): Promise<void> {
+    var contentDisposition = require('content-disposition')
```
bpatrik, line 87: please fix the lint errors.

src/backend/middlewares/GalleryMWs.ts
```diff
@@ -102,7 +103,16 @@ export class GalleryMWs {
 
     try {
       res.set('Content-Type', 'application/zip');
-      res.set('Content-Disposition', 'attachment; filename=Gallery.zip');
+      var zipName = directoryName;
```
bpatrik, line 106: please fix the lint errors.

src/backend/middlewares/GalleryMWs.ts
bpatrik, line 87: Is this a new dependency to the project? It is not listed in the package.json
Can you reason about why this is needed?

bpatrik (changes requested):

src/backend/middlewares/GalleryMWs.ts
fred1701d: I don't know how to fix it. I can decline my PR
I'm not a Node JS developer. Java is much better :-)

fred1701d: I don't know, that's something I found online.
I think I'll just drop my PR

Issue #794, opened 2023-12-01
Special characters displayed wrongly


grasdk:
**Describe the bug**

Danish letters Ææ, Øø and Åå and German ü - and probably other characters are not displayed correctly in the user interface when saved as "Region Person Display Name" or "Region Name", not sure which one is actually read. The metadata is added by DigiKam 8.1.0, but as far as I can gather, it is stored as UTF-8.

See screenshot and attached photo (borrowed from [wikimedia](https://commons.wikimedia.org/wiki/File:People_are_walking_and_standing_in_the_street_Zeedijk_in_Amsterdam_old_city;_free_download_photo_in_high_resolutions_by_Fons_Heijnbroek,_street_photography_The_Netherlands,_CCO.tif))

exiftool output:
`$ exiftool -codedcharacterset 2023-08-27-120000-example.jpg`
`Coded Character Set             : UTF8`

`$ exiftool -Region* -Keywords -XP* 2023-08-27-120000-example.jpg`
`Region Person Display Name      : Person Carrying ChildInYellowDress, Pærsøn Åkessün Æñtestå, RedHairedPerson SittingOnBench`
`Region Rectangle                : 0.611242, 0.453747, 0.0214017, 0.0391972, 0.226011, 0.500157, 0.019285, 0.0319849, 0.365945, 0.486359, 0.00470367, 0.00940734`
`Region Applied To Dimensions H  : 3189`
`Region Applied To Dimensions Unit: pixel`
`Region Applied To Dimensions W  : 4252`
`Region Area H                   : 0.0391972, 0.0319849, 0.00940734`
`Region Area Unit                : normalized, normalized, normalized`
`Region Area W                   : 0.0214017, 0.019285, 0.00470367`
`Region Area X                   : 0.621943, 0.235654, 0.368297`
`Region Area Y                   : 0.473346, 0.516149, 0.491063`
`Region Name                     : Person Carrying ChildInYellowDress, Pærsøn Åkessün Æñtestå, RedHairedPerson SittingOnBench`
`Region Type                     : Face, Face, Face`
`Keywords                        : Holiday, RedHairedPerson SittingOnBench, Person Carrying ChildInYellowDress, Pærsøn Åkessün Æñtestå`
`XP Keywords                     : Holiday;RedHairedPerson SittingOnBench;Person Carrying ChildInYellowDress;Pærsøn Åkessün Æñtestå`

**Photo/video (optional) that causes the bug**

[...]

PR #826, opened 2024-02-05
Fix for issue #794 (bugfix)

grasdk:
* Switched from ExifReader to exifr and updated tests and test material to fix issue [#794](https://github.com/bpatrik/pigallery2/issues/794)
* Removed dependency to ExifReader
* Updated .gitignore
* Minor clarification added to CONTRIBUTING.md


grasdk:
While doing this, I discovered a bug with png files having the "createDate" set to the file-date, even if they have metadata embedded. Exifr can read this date metadata from png files, so there are options to further utilize exifr for fixing issues.

I will look into this, but wanted to separate it from the fix of the issue.

.gitignore
```diff
@@ -34,3 +34,11 @@ test.*
 *.sublime-project
 *.sublime-workspace
 .DS_Store
+test/folder-reset.js
```
bpatrik, line 37: Can you just use wildcard instead, like: test/*.js etc?

bpatrik (changes requested):
Thank you! Nice contribution!


Can you please also remove any unused dependencies from the package.json (and also from the lock file)

grasdk:
> Thank you! Nice contribution!
> 
> Can you please also remove any unused dependencies from the package.json (and also from the lock file)

Sure. I will do this asap. Hopefully tonight (it's morning here, now). Also the .gitignore

bpatrik:
Thank you!

Issue #808, opened 2024-01-06
Playing videos results in "Error during loading the video."


drwb:
**Describe the bug**

Playing videos has stopped working, resulting in a "Error during loading the video." 

It's been working perfectly for months.

![SCR-20240106-jwr](https://github.com/bpatrik/pigallery2/assets/16479230/5dff2e23-bc4b-44a6-ab05-0697288ef943)

**Photo/video (optional) that causes the bug**
The mediainfo for one of the videos is below, but it is repeatable for each video in each folder.
<details><summary>mediainfo  1.mp4</summary>
<p>

General
Complete name                            : 1.mp4
Format                                   : MPEG-4
Format profile                           : Base Media
Codec ID                                 : isom (isom/iso2/avc1/mp41)
File size                                : 809 MiB
Duration                                 : 8 min 2 s
Overall bit rate                         : 14.1 Mb/s
Writing application                      : Lavf57.71.100

Video
ID                                       : 2
Format                                   : AVC
Format/Info                              : Advanced Video Codec
Format profile                           : Main@L3.1
Format settings                          : CABAC / 4 Ref Frames
Format settings, CABAC                   : Yes
Format settings, Reference frames        : 4 frames
Codec ID                                 : avc1
Codec ID/Info                            : Advanced Video Coding
Duration                                 : 8 min 2 s
Bit rate                                 : 14.0 Mb/s
Width                                    : 1 080 pixels
Height                                   : 1 920 pixels
Display aspect ratio                     : 0.562
Frame rate mode                          : Constant
Frame rate                               : 30.000 FPS
Color space                              : YUV
Chroma subsampling                       : 4:2:0
Bit depth                                : 8 bits
Scan type                                : Progressive
Bits/(Pixel*Frame)                       : 0.225
Stream size                              : 799 MiB (99%)
Writing library                          : x264 core 148
[...]

PR #827, opened 2024-02-06
Fix video not loading #808

BluSyn:
Route collision with thumbnail path and video path. Results in videos not loading properly.

Changing the route definition order fixes the issue.


bpatrik:
Thank you!

PR #829, opened 2024-02-11
consolidate exif parsing libraries - rework of timestamps

grasdk:
**Metadata Parsing**
- metadata.CreatedDate now supported for png.
- Switched to exifr to parse timestamps and timezone offset
- Switched to exifr for camerainfo
- Removed ExifParser library dependency
- Always using image-size to get image-dimensions, but has fallback to ifd0 and exif dimensions in case this fails.
-- Rather than getting rid of image-size, it appears to be a good idea to use an independent source for image dimensions and only use tags as fallback (judging from a thread on exiftool's forums most software do not tag the dimensions of an image correctly).

I believe this comes pretty close to resolving https://github.com/bpatrik/pigallery2/issues/277

**Timestamp handling**
metadata.CreatedDate is now milliseconds based on UTC, where enough information was available. The value is read from the most accurate tag that is available. Preceedence is

1. DateTimeOriginal (camera shutter close time)
2. CreateDate (camera file creation time (typically same time as camera shutter close time)
3. ModifyDate (as the name indicates a date of modification which is why it has lower preceedence)
4. FileDate (filesystem date - is often updated by just copying a file or cloning from git, so worst timestamp of them all)

Offsets are pulled directly from the metadata if availalbe. Same preceedence as above. If no offset is available, but GPS is, the offset is calculated from GPS timestamp (which is always UTC by spec). If no offset is available by metadata or calculation, offset is not defined.

- updated all test cases to reflect these timestamp changes
- Added several test-cases for timestamps.

*Examples*
These changes mean that a picture taken in London and a picture taken in Sydney at the exact same millisecond will have the same CreatedDate value.
The offset (e.g. +00:00 or +01:00 for London and +10:00 or +11:00 for Sydney) has been added to the photometadata, where available. This can be used to give the user better information in the UI.


grasdk:
*I realize that this change may impact users of pigallery who live far away from UTC. Without this change, dates are parsed as local time on the server that runs pigallery, that is how Javascript Date works. To amend it, I have made the offset time available in the metadata, but it has not yet been taken into used. Until the time that the UI can utilize this offset, this change will cause a daytime photo taken in e.g. Sydney to show up with a night time (or evening or morning) timestamp in Europe and vice versa. However for the overall correctness, I believe that this approach is correct.*

grasdk:
> _I realize that the timestamp handling change may impact users of pigallery who live far away from UTC. Without this change, dates are parsed as local time on the server that runs pigallery, that is how Javascript Date works. To amend it, I have made the offset time available in the metadata, but it has not yet been used. Until the time that the UI can utilize this offset, this change will cause a daytime photo taken in e.g. Sydney to show up with a night time (or evening or morning) timestamp in Europe and vice versa. However for the overall correctness, I believe that this approach is correct._

Most of these changes have now been mitigated with updates to the UI. Offset will be displayed and taken into account when available.

One caveat: Search using before and after date will occasionally miss a picture, because search is done by UTC time. So a new-year photo from Australia early the 1st of January, will technically still be 31st of December UTC, so it won't show up when using after: and first of January.

grasdk:
By the way. The photos added for demo and test are from https://unsplash.com/ and free to use for any purpose except reselling or competing with unsplash.

grasdk:
Hey @bpatrik

When you get to review this, which is a combined metadataloader refactor and use of timestamps with offsets. You may want to take a look at this small additional change, which I have ready for merge aswell, if you like it:

https://github.com/grasdk/pigallery2/compare/master...grasdk:pigallery2:feature/location_info?diff=split&w=

In the lightbox and infopanel for photos, it will show all of the location metadata: city, state and country separated by comma. Any empty value will be left out of course.

My reasoning for suggesting this change is programs like GeoSetter and GetTagNinja fill out these values, so you can display "Los Angeles, California, USA" or "Birmingham, England, Great Britain" for example. Many countries don't use the "State" attribute, it there it will just be like "Athens, Greece" for example.

I can make it a part of this pull request or make another pull request if you think it's a good idea.

No rush ;)

Kind regards.

src/common/entities/PhotoDTO.ts
```diff
@@ -33,6 +33,7 @@ export interface PhotoMetadata extends MediaMetadata {
   positionData?: PositionMetaData;
   size: MediaDimension;
   creationDate: number;
+  creationDateOffset?: string;
```
bpatrik, line 36: There is a ContentWrapper class that does some "manual compression" over the data, where it replaces the lengthy json names with a single character. Here: https://github.com/bpatrik/pigallery2/blob/7fa2594f762a4b24443b2ddd0d83772aa1aee63c/src/common/entities/ConentWrapper.ts#L79-L80

It also has a pair for decompression in the same file. 
Can you please add this new property there too. Can you also take a look if this new data can be better represented in the DB? like converting it to a int/tnyint. Would that work or on the client side a lot of extra back conversion would be needed? 


Some quick reasoning why we have the "manual compression": At some point I totally though that nginx gzips josn data and the amount data that the app transferred was enormous still. So I implemented this manual compression. That helped a lot with large search results and folders. Then I actually set up nginx to compress that improved even more. 
Until the need of the manual compression is tested with proper gzipping, I would like to keep the code consistent. (Also protobuf would be probably the best solution here)

bpatrik:
> > _I realize that the timestamp handling change may impact users of pigallery who live far away from UTC. Without this change, dates are parsed as local time on the server that runs pigallery, that is how Javascript Date works. To amend it, I have made the offset time available in the metadata, but it has not yet been used. Until the time that the UI can utilize this offset, this change will cause a daytime photo taken in e.g. Sydney to show up with a night time (or evening or morning) timestamp in Europe and vice versa. However for the overall correctness, I believe that this approach is correct._
> 
> Most of these changes have now been mitigated with updates to the UI. Offset will be displayed and taken into account when available.
> 
> One caveat: Search using before and after date will occasionally miss a picture, because search is done by UTC time. So a new-year photo from Australia early the 1st of January, will technically still be 31st of December UTC, so it won't show up when using after: and first of January.

Just to double check. After these change, this means that If I take a photo at noon (12:00) in London and someone else takes a photo 5 hours later at noon (12:00) in New York. Both of them would be shown in the app as they would have been created at the same time at noon. Even if I watch the photos in Sydney time. 

That would be the expected behavior as when you look back to your holidays photos, you do not expect to see a lunch photo taken at 2am. 

One future feature I can imagen is to add a switch to the lightbox menu that switches between local time and UTC (or however we would call it)

src/frontend/app/ui/gallery/navigator/sorting.service.ts
bpatrik, line 1: Can you also make sure both grouping and sorting takes the new dateoffset into account.

src/frontend/app/ui/gallery/filter/filter.service.ts
```diff
@@ -205,7 +205,7 @@ export class FilterService {
     const startMediaDate = new Date(floorDate(minDate));
 
     prefiltered.media.forEach(m => {
-      const key = Math.floor((floorDate(m.metadata.creationDate) - startMediaDate.getTime()) / 1000 / usedDiv);
+      const key = Math.floor((floorDate(m.metadata.creationDate) - startMediaDate.getTime()) / 1000 / usedDiv); //TODO
```
bpatrik, line 208: please finish.

Maybe consider storing offset as int (sec/millisec) so it would be ease to add here. Also use in the SQL expression for search.

bpatrik (changes requested):
I think its fine if the search update is not part of this PR. 

But can you at least add a \\TODO with some explanation to the right places in the file.
One sample place: https://github.com/bpatrik/pigallery2/blob/7fa2594f762a4b24443b2ddd0d83772aa1aee63c/src/backend/model/database/SearchManager.ts#L566

bpatrik:
Also thank you very much. This is huge contribution to the app!

src/common/entities/PhotoDTO.ts
grasdk: Of course. I will do this ;)

Thank you for the reference and background. Sometimes knowledge of data allows you to optimize in ways generic optimizations can't do. :)

src/frontend/app/ui/gallery/navigator/sorting.service.ts
grasdk: Sure, I can do something. If nothing else add a comment about it. 

The reason why I didn't do anything is that in the sense of universal time, sorting and grouping already works. I thought it would probably be what was to be expected, but let me know if you think otherwise.

Here's an example: Let's say Alice and Bob have a phone call from London (UTC+0) to Amsterdam (UTC+1) respectively. Alice asks Bob to take a selfie with him and an Amsterdam landmark. He takes the photo at 11:30 local time in Amsterdam with his phone that adds the GPS, timezone and everything to the photo metadata. Bob then asks Alice to reciprocate, and Alice takes a selfie in London 10:31 local time.

They later get married and pool their photos together in their self hosted NAS using Pigallary2 to view and share. They would expect (I think) these pictures to show up in the order they were taken. :)

Chronologically, Bob's photo was taken first at 11:30 UTC+1, then Alices at 10:31 UTC+0.

With the current date handling, this will happen without further modifications to the grouping and sorting as you can see from the simple Javascript test below:
```
new Date("2024-02-21 11:30:00+01:00").getTime(); //1708511400000
new Date("2024-02-21 10:31:00+00:00").getTime(); //1708511460000
```
The 10:31 UTC+0 time is 60000 ms after 11:30 UTC+1 time. And these are the values in use by Pigallery2.

We could probably do a modification so 10:31 London time is before 11:30 Amsterdam time. Let me know if we should and I will try to figure a good way to do it.

Cheers ;)

grasdk:
> > > _I realize that the timestamp handling change may impact users of pigallery who live far away from UTC. Without this change, dates are parsed as local time on the server that runs pigallery, that is how Javascript Date works. To amend it, I have made the offset time available in the metadata, but it has not yet been used. Until the time that the UI can utilize this offset, this change will cause a daytime photo taken in e.g. Sydney to show up with a night time (or evening or morning) timestamp in Europe and vice versa. However for the overall correctness, I believe that this approach is correct._
> > 
> > 
> > Most of these changes have now been mitigated with updates to the UI. Offset will be displayed and taken into account when available.
> > One caveat: Search using before and after date will occasionally miss a picture, because search is done by UTC time. So a new-year photo from Australia early the 1st of January, will technically still be 31st of December UTC, so it won't show up when using after: and first of January.
> 
> Just to double check. After these change, this means that If I take a photo at noon (12:00) in London and someone else takes a photo 5 hours later at noon (12:00) in New York. Both of them would be shown in the app as they would have been created at the same time at noon. Even if I watch the photos in Sydney time.
> 
> That would be the expected behavior as when you look back to your holidays photos, you do not expect to see a lunch photo taken at 2am.
> 
> One future feature I can imagen is to add a switch to the lightbox menu that switches between local time and UTC (or however we would call it)

Yes. Lightbox etc. where time is displayed for human viewing, will show 12:00Z in London and 12:00-05:00 for New York (the default for Datepipe formatting is that UTC+00:00 is displayed at Z). This is regardless of where Pigallery is running.

Only when absolutely no information about offset is available, the local time of the pigallery2 server will be used. So if your pigallery is running in timezone +1 and you show a picture from New York without offset. It would probably show the time wrong by 6 hours (the difference between the -5 and +1 offset). That is how it works now without this pull request, if my experiments are correct. Most modern photos won't have this problem, but incomplete metadata from old cameras, scanned photos may not have saved any timezone info. But I would suggest to the owner of such photos to fix this using exiftool or one of the other tools available for this :)

src/frontend/app/ui/gallery/filter/filter.service.ts
grasdk: I'm looking at this and perhaps the argument from above, with sorting applies here aswell, what do you think?

src/common/entities/PhotoDTO.ts
grasdk: https://github.com/bpatrik/pigallery2/pull/829/files#diff-c2af469ef10300930189ed2851d98d0efbbf2620572cd4c81e9cbbd413d7197dR82-R86

https://github.com/bpatrik/pigallery2/pull/829/files#diff-c2af469ef10300930189ed2851d98d0efbbf2620572cd4c81e9cbbd413d7197dR346-R353

src/frontend/app/ui/gallery/navigator/sorting.service.ts
grasdk: I'll mark as resolved with the words above. Let me know if you disagree ;)

bpatrik:
Great! Thank you very much for your contribution!  I think I can go ahead to merge it.

bpatrik:
ahh this PR ended up breaking tests :( https://github.com/bpatrik/pigallery2/actions/runs/8125512654/job/22208202972

for `search date pattern` tests

grasdk:
Ok. I will take a look asap. I will prioritize this as much as possible.

grasdk:
Hi @bpatrik 
I finally figured out what the problem is. It's leap year and we're past the 29th of february. The tests that fail compare day of year this year with day of year other years. Today (march 5th) is 65th day of 2024, but only 64th day of year 2023. Hence we have some test data that isn't properly picked up :)

It also explains why I didn't get the error in the pull request build, because it was before the 29th of february.

I'm thinking that replacing the format string "%j" (dayofyear) with "%m%d" (month and date) will help. I will try this out and report back (hopefully with a pull request that has a fix)

bpatrik:
I'm sorry, but apparently there were not enough tests on metadataloader and the timestamps for some photos are borken now in the demo folder. These photos have a proper date:

kép

bpatrik:
Here is a quick var console.log dump. I will let this with you as I suspect the issue is with the new exifr part:

Input file: wild-1.jpg from the demo folder.
```
stat.mtime.getTime: 1694635419042
iptcData.date_time.getTime: 1435917883000
exif.exif: {                                                   
  ExifVersion: Uint8Array(4) [ 48, 50, 51, 49 ],    
  CreateDate: '2015:07:03 10:04:43',                
  OffsetTime: '+02:00',                             
  SubSecTimeDigitized: '50',                        
  ColorSpace: 1,                                    
  ImageUniqueID: '5256C01EA869151BFA5D760D7D06D6B1',
  SerialNumber: '123063022888',                     
  LensInfo: [ 15, 85, NaN, NaN ],                   
  LensModel: 'EF-S15-85mm f/3.5-5.6 IS USM',        
  LensSerialNumber: '0000129324'                    
}       
```
```
output:                                            
{                                                                                
  size: { width: 1224, height: 816 },                                            
  creationDate: 0,                                                               
  fileSize: 632861,                                                              
  positionData: { country: 'United States', state: 'Arizona', city: 'Williams' },
  keywords: [ 'Bearizona', 'USA', 'USA Road trip' ],                             
  cameraData: {                                                                  
    make: 'Canon',                                                               
    model: 'Canon EOS 600D',                                                     
    lens: 'EF-S15-85mm f/3.5-5.6 IS USM'                                         
  },
  creationDateOffset: '+02:00',
  faces: [
    { name: 'Balu the bear', box: [Object] },
    { name: 'Hugin the raven', box: [Object] },
    { name: 'Akela the wolf', box: [Object] }
  ]
}

```

The 'solution'  should be `'2015:07:03 10:04:43`.

bpatrik:
Maybe this should be `exif.exif.CreateDate`?
https://github.com/bpatrik/pigallery2/blob/3811fc3c998e3b0f20dc3c65c552cc2f9e53a006/src/backend/model/fileaccess/MetadataLoader.ts#L380
pls doublecheck.

Issue #609, opened 2023-02-13
Support for Apple Live Photos


FartherOut:
I have lots of live photos taken with various Apple devices, but the gallery displays the photo and the accompanying video as two separate objects. This makes the gallery rather cluttered and inelegant.

A live photo is basically a HEIF/JPG and a MOV/MP4 with the same filename.
Having support for live photos would be tremendous.

bpatrik:
Duplicate of #97

FartherOut:
Oh.
Live photos support, which I desire, and Heic support are two different things though.
Live photos does not have to be in a heic format.
You basically group together two files with the same filename but different file ending.

There are metadata which you can use to identify a live photo, such as `MediaGroupUUID` in the photo and `ContentIdentifier` in the movie, but I am unsure if you actually need this.

bpatrik:
Ah sorry I misunderstood.

bpatrik:
@martadinata666 made some change. maybe it works now. (testing needed)

Issue #97, opened 2019-08-14
Request: (Apple) HEIC images


iAmRenzo:
Support of heic images from iPhone

bpatrik:
Hi 

unfortunately non of the browsers are supporting that file format: 
https://caniuse.com/#feat=heif
So I cannot just add it now.

In the future I'll implement a *.* to *.jpg converter in the app, with that feature I can support heic too. (sorry deadline promises, probably will take months)

Patrik

bpatrik:
Hi,

unfortunately its not easy to add ass HEIC not an open format. Read more here: https://github.com/lovell/sharp/issues/1105

So I cannot add this in the near future.

Best,
Patrik

helmut72:
lovell/sharp#1105 is closed

bpatrik:
A gave it a quick look (with sharp v0.27.0 and I still get `heif: Unsupported feature: Unsupported codec (4.3000)`
(Tested on https://github.com/tigranbs/test-heic-images/blob/master/image1.heic).

Without looking in to it in depth, this: https://github.com/lovell/sharp/issues/1105#issuecomment-631143395 suggests that works only on mac?

presto8:
I am not that familiar with Node development, but would it be possible to use something like heif-convert (external command-line tool, does not require a Mac). If heif-convert could be used to transcode HEIF to JPG and then store the JPG files in the cache/thumbnails directory, it would be a huge benefit for those of us who have huge HEIF libraries. Thank you!

presto8:
More info on this: libvips supports HEIF now (tested on Linux, version 8.12.1). Since sharp uses libvips, I think this feature may work if the sharp dependency is updated to the latest version?

bpatrik:
There is an open issue about upgrading sharp to the latest version: https://github.com/bpatrik/pigallery2/issues/295

Unfortunately it is not trivial, as I would like to keep supporting arm based Docker images

yllekz:
If it helps, "viktorstrate/PhotoView" is capable of reading the HEIC format and it runs on Pis/M1/etc. Might be able to have a look at his implementation.

bpatrik:
There was a sharp update recently, so this request might be  unblocked. Someone need to check it.

bpatrik:
try adding it here and see what happens: https://github.com/bpatrik/pigallery2/blob/1e8830ab09686da171b2a0a49fec5eac79943646/src/common/SupportedFormats.ts#L2

[...]

PR #833, opened 2024-02-20
Manually add vips-heif

martadinata666:
Unlike debian that pulled heif libs when install vips-dev, alpine need to install this heif separately.


bpatrik:
referencing: #609 #561 #97

martadinata666:
Oh, yah that time we don't touch debian base yet, as we consider this something experimental thing. I will send PR on debian later.

PR #834, opened 2024-02-21
Add heif support via VIPS (debian docker image)

martadinata666:
Building sharp module with respective distro vips libs.


bpatrik:
Thank you!

Issue #835, opened 2024-02-22
Scroll through images using horizontal scroll action


veroxzik:
**Is your feature request related to a problem? Please describe.**
The behavior of the horizontal scroll (i.e. tilting scroll wheel on mouse left and right) produces an unusual behavior. When viewing a single image, tilting in either direction will zoom the photo out. I don't believe it does any other action elsewhere.

**Describe the solution you'd like**
I would like horizontal scroll to move left and right through the images, as if you pressed the arrow keys or clicked the arrows on the left or right.

---

Thank you for the wonderful project! 
I am not very familiar with typescript, but if you could give a pointer on where to look, I could attempt a PR to address this.

PR #836, opened 2024-02-23
Lightbox: fix mouse wheel behavior and add horizontal scroll feature.

veroxzik:
This addresses my issue #835 and is two fold:
1. Previously, any non-negative deltaY change to the mouse wheel would cause a photo in the lightbox zoom out, regardless of the direction. I have changed this so that zoom out is only possible with positive deltaY values.
2. I have added the option to scroll through photos in the lightbox using horizontal scrolling (deltaX). This mimics the behavior of left/right arrow and allows for navigation solely with the mouse. I've also changed it so you _can_ use this scroll on videos. The previous behavior of canceling zoom on videos remains.

Thanks for the consideration!


bpatrik:
Great! Thank you for your contribution!

PR #843, opened 2024-03-03
Add logging elision

mblythe86:
Allow for anonymous functions in logging calls.
The function is only called if the message is logged. (e.g. if the verbosity is turned up high enough.)
This allows for expensive operations to be avoided in cases where the logging won't happen.

The idea is that this provides a performance benefit. I don't know how "expensive" an operation must be
to actually realize a performance benefit, though.

I added an example in `server.ts`...
It's probably "expensive" to dump the configuration to JSON, then stringify that JSON for logging.

I originally implemented something similar in my 'digikam-connector' extension, where I used it to log the inputs/outputs of the extension hooks while I was developing and debugging.  I figured that running `util.inspect()` on everything was undesirable in a production environment, but didn't want to remove all my logging statements.  So, this seemed like a good compromise.

In my local testing, the typescript still compiles OK, and the messages from `server.ts` print out exactly as they did before.

### DIFF
--- a/src/backend/Logger.ts
+++ b/src/backend/Logger.ts
@@ -1,7 +1,7 @@
 import {Config} from '../common/config/private/Config';
 import {LogLevel} from '../common/config/private/PrivateConfig';
 
-export type logFN = (...args: (string | number)[]) => void;
+export type logFN = (...args: (string | number | (() => string))[]) => void;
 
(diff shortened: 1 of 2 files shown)


bpatrik:
Sounds good to me!

I don't think it matters much tho.

PR #844, opened 2024-03-05
Add excludeDir to extension interface

mblythe86:
Without this, the mechanism to exclude a directory from indexing is a bit strange.  I had to use the `scanDirectory.after` hook, then remove elements from the `output.directories` array.

In order to make the extension hook a bit more developer-friendly, I also changed the arguments to the `excludeDir` function to be a dictionary instead of just an array (or tuple?) of strings.

Tested and working against commit f5f34c5 with a tweaked version of my 'digikam connector' extension.


bpatrik:
Cool, thanks!

PR #850, opened 2024-03-07
Feature/location info (#9)

grasdk:
* Show comma-separated city, state and country in infobox and in gallery mode. All of the values that are non-empty will be shown.

Before:
image

After:
image


grasdk:
A matter of taste perhaps. I thought with the nice searchable data, it might also be displayed. Especially in places like the USA where the state is rather important info. For example: Beverly Hills exist in 5 different states, not only the one in California :)

bpatrik:
Thanks!

PR #860, opened 2024-03-23
Deal with rating missing in XMP sidecars

martyone:
Avoid getting Rating of 0 when rating metadata is not included in a sidecar.

### DIFF
--- a/src/backend/model/fileaccess/MetadataLoader.ts
+++ b/src/backend/model/fileaccess/MetadataLoader.ts
@@ -175,7 +175,10 @@ export class MetadataLoader {
                 delete (sidecarData as any)['xap'];
               }
               if ((sidecarData as SideCar).xmp !== undefined) {
-                if ((sidecarData as SideCar).xmp.Rating !== undefined) {
+                if (
+                    (sidecarData as SideCar).xmp.Rating !== undefined &&
+                    (sidecarData as SideCar).xmp.Rating > 0
+                ) {
                   metadata.rating = (sidecarData as SideCar).xmp.Rating;
                 }
                 if (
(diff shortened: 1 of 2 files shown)


grasdk:
I found the reverse bug in the embedded metadata handling. Here a rating of zero will be ignored. I have a fix for it, but it will be some days before it's ready because it's bundled with some other minor improvements of metadataloader and I'm not able to finish it in the next few days.

grasdk:
@martyone looking at this again, and your corrections to the code. 

With this code, if rating is defined, but zero, it is not mapped it into pigallery's metadata. Is this the intention?

I ask because both the jpg and the xmp-file has a rating of 0, but your test is altered to expect no mapped rating:
https://github.com/bpatrik/pigallery2/pull/860/commits/8517430a1b48b3570e0383409dc5c28fa61f8797#diff-c05a31981597794f928cda5f1343aadd8c148f455efbab3d572210c6801d90eeL14

But rating is defined to be zero:
https://github.com/martyone/pigallery2/blob/xmp-no-rating/test/backend/assets/sidecar/metadata_v2.xmp#L8

(also in the JPG):
```
$ exiftool -Rating* test/backend/assets/sidecar/metadata_v2.jpg
Rating Percent                  : 0
Rating                          : 0
```

If you look at [no_metadata_v3.xmp](https://github.com/martyone/pigallery2/blob/xmp-no-rating/test/backend/assets/sidecar/no_metadata_v3.xmp) on the other hand, there is no rating defined at all. In such a case, I think no rating should be expected, but with this merge, rating = 0 is effectively ignored.


----

I'm working on my own fork, pulling in these changes. My work is about consolidating all this metadata mapping. Currently there are 3 copies of roughly the same logic:
embedded metadata, video-sidecar and photo-sidecar

I want to make them into the same logic, taking all old and new changes and functionality into account of course. I'm a bit troubled by this one, though. Hence the question above. Is it intentional to ignore rating 0 and if so, why? :)

kind regards :)

martyone:
@grasdk yes, it was intentional to ignore rating = 0. I decided based on my experience with Digikam. When you set rating and later unset it, Digikam stores rating = 0 in metadata instead of clearing the property. So Digikam makes no distinction between no rating and rating = 0 AFAICT. In PG2 I originally got a mix of rating=<unknown> and rating=0 in that case, which just cluttered the filter.

grasdk:
Love to Digikam :) I use it aswell. Pigallery2 is a perfect fit to "share" your metadata work in digikam in a useful (and fast) way :)

PR #862, opened 2024-03-29
Trim extensions when sorting filenames

### LINKED ISSUE #518: Alphabetical sort isn't correct for images like 10 and 10.1
**Describe the bug**

I have a directory with images, and they are just named by numbers. These numbers are the correct order in which the images should be displayed, but some have a .1 or further to show they are other versions of the actual image.  The correct order would be:

- 10
- 10.1
- 10.2

But it shows up as
- 10.1
- 10.2
- 10

**Photo/video (optional) that causes the bug**

These images are quite private, so I'd rather not upload them here.

 - OS: Running in docker on an Ubuntu system
 - Browser: Edge

**Used app version**:
- docker-latest

sarayourfriend:
Fixes #518

I've described the root cause in this comment on the issue: https://github.com/bpatrik/pigallery2/issues/518#issuecomment-2026473493

To test:
- Checkout the branch and rename some of the demo images to `10`, `10.1`, `10.2`, etc.
- Run the app and confirm that they now sort as expected according to the issue (10, 10.1, 10.2 rather than 10.1, 10.2, 10).
- Download a PNG ([here's a small filesized CC0 one from Wikimedia Commons you can use](https://commons.wikimedia.org/w/index.php?curid=931234) and name it 10.png in the demo images folder. Reverse the name sorting a couple times and confirm that 10.jpg and 10.png correctly trade places and sort alphabetically by their full filename. This test is crucial to confirm consistency in the name output, otherwise the identical extensionless names would make sorting them non-deterministic.

### DIFF
--- a/src/common/Utils.ts
+++ b/src/common/Utils.ts
@@ -173,7 +173,7 @@ export class Utils {
   }
 
   static getOffsetMinutes(offsetString: string) { //Convert offset string (+HH:MM or -HH:MM) into a minute value
-    const regex = /^([+-](0[0-9]|1[0-4]):[0-5][0-9])$/;  //checks if offset is between -14:00 and +14:00. 
+    const regex = /^([+-](0[0-9]|1[0-4]):[0-5][0-9])$/;  //checks if offset is between -14:00 and +14:00.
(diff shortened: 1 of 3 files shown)


bpatrik:
Thank you for the contribution!

Issue #519, opened 2022-07-08
Output loging attempts so i can use fail2ban to ban bruteforces?


kardo22:
**Is your feature request related to a problem? Please describe.**
I am using pigallery behind a reverse proxy, i have been looking at the logs but i don't think the reverse proxy can determine if the login attempt was failed or successful, it simply just shows the client fetching resources from the webpage and if its a fail it reloads the page.

**Describe the solution you'd like**
A clear output of logging directly from pigallery docker container that i can read from with fail2ban so i can use it to protect pigallery against bruteforce attacks.

**Describe alternatives you've considered** (optional)
Ive been looking around inside the docker container, i expected the log to reside in /var/log/ but there is nothing there. In case there is already a log file, can you please let me know where it is found inside the container?

bpatrik:
related to #214

bpatrik:
I was also looking into this. 

(Un)fortunatly my app is not famous enough for targeted attacks from botnets. What I saw in logs that they try if they can use the default admin admin password to get into a worldpress app. 

What I come up (but yet to implement) is to move the `/api` to `/pgapi` and ban all calls to `/api` in the nginix config.

Also login should be a bit more clean agree.

Docker has its own logging. My app only prints logs to stdout. So you can find the logs if you do docker ps xzy

PR #863, opened 2024-03-29
Use separate log levels for relevant error types

### LINKED ISSUE #514: 'Invalid CSRF token'
**Describe the bug**

Just installed from the docker image, getting this error constantly, both in the log, and as a red popup. I have put the service behind a reverse proxy and am using SSL.

```
7/3/2022, 7:06:46 AM[WARN_] Handled error:
ErrorDTO {
  code: 1,
  message: 'Invalid CSRF token',
  details: ForbiddenError: invalid csrf token
      at csrf (/app/node_modules/csurf/index.js:112:19)
      at result (/app/node_modules/express-unless/index.js:49:5)
      at processTicksAndRejections (internal/process/task_queues.js:95:5) {
    code: 'EBADCSRFTOKEN'
  },
  request: { method: 'POST', url: '/api/graphql' },
  detailsStr: 'ForbiddenError: invalid csrf token'

```

sarayourfriend:
Potentially fixes #514 (depending on interpretation of what would resolve that issue)

I wrote a [little in the issue of my exploration of this problem](https://github.com/bpatrik/pigallery2/issues/514#issuecomment-2026765422). Basically, I think the issue comes down to a matter of opinion about what should and shouldn't be logged at the "warning" level.

In my experience, authentication errors are not worth logging at all in production environments, precisely because they happen so frequently with credentials stuffing or bots crawling for exposed WordPress/Drupal/Django/whatever admin pages. It is trivial to detect these based on request logs and response status code volumes, and even from pigallery2's own access logs, if enabled (though, those are debug-level so are a different question).

I admit that is an opinion, but for well configured servers, these logs will either unnecessarily frighten people who don't have experience running application services in production environments or fill up your log files and potentially cost you additional money in your log querying/transport solution if you have one. I run applications at a warn level in production, and wouldn't expect that to include anything other than predicted-but-unusual/notable/potentially-bad cases (contrasted to errors which would be unhandled edge cases/bugs).

To reflect that, this PR:

1. Defines a log level for each `ErrorCodes` value, or at least defines specific ones for several error types and defaults to warn. This has an added benefit of actually increasing the log level for what may be true error-level stuff, though that's definitely something that needs input. I'm not familiar with all the features of pigallery2 as I only use the simplest ones myself, and haven't run into most of these edge cases to say whether they're definitely errors or warnings. This is easy to change though, just change which set of cases any particular error code is grouped into.
2. Expands `LoggerArgs` to accept simple objects and `Error` instances. These get forwarded to `console.log` in `Logger.log` in the end, which is able to serialise such arguments and print them nicely. It looks like sometimes the `console` functions are used directly instead of `Logger` ones to leverage it's nice printing of objects, but the differences between `Logger.error` and `console.error` make it difficult to quickly update everywhere to use `Logger`. That could be done more carefully in a separate PR if desired. 
3. Updates the `RenderingMWs::renderError` function to use the log-level defined for the rendering error's error-code. This is the effective change which lowers authentication errors to debug level logs, removing them from production logs under the default logging configuration.

I also found some dead code in the rendering middleware's error logging routine and removed it. I'll add a comment explaining why it is dead code to clarify.

Also, I want to clarify again that this is just my interpretation of what the appropriate logging levels are for these kinds of authentication errors, based on my experience developing and running several different internet-facing applications in production. Happy to be wrong about this for pigallery2! Not all applications have the same logging needs.

I look forward to any comments on the PR :slightly_smiling_face: I really like pigallery2 and think it's a great application, and am excited to contribute any way I can.

### DIFF
--- a/src/backend/Logger.ts
+++ b/src/backend/Logger.ts
@@ -1,7 +1,6 @@
 import {Config} from '../common/config/private/Config';
 import {LogLevel} from '../common/config/private/PrivateConfig';
 
-export type logFN = (...args: (string | number | (() => string))[]) => void;
 
 const forcedDebug = process.env['NODE_ENV'] === 'debug';
 
@@ -11,7 +10,8 @@ if (forcedDebug === true) {
   );
 }
 
-export type LoggerArgs = (string | number | (() => string))
+// Match `console.log` inputs
+export type LoggerArgs = Parameters<Console["log"]>[0]
 export type LoggerFunction = (...args: LoggerArgs[]) => void;
 
 export interface ILogger {
--- a/src/backend/Logger.ts
+++ b/src/backend/Logger.ts
@@ -1,5 +1,6 @@
 import {Config} from '../common/config/private/Config';
 import {LogLevel} from '../common/config/private/PrivateConfig';
+import {ErrorCodes} from '../common/entities/Error';
 
 
 const forcedDebug = process.env['NODE_ENV'] === 'debug';
@@ -102,5 +103,26 @@ export class Logger {
       }
     });
     console.log(date + tag + LOG_TAG, ...args);
+  }
+
+  public static logLevelForError(e: ErrorCodes): LoggerFunction {
+    switch (e) {
+      case ErrorCodes.INPUT_ERROR:
+      case ErrorCodes.NOT_AUTHENTICATED:
+      case ErrorCodes.ALREADY_AUTHENTICATED:
+      case ErrorCodes.NOT_AUTHORISED:
+      case ErrorCodes.PERMISSION_DENIED:
+      case ErrorCodes.CREDENTIAL_NOT_FOUND:
+        return Logger.debug
+      case ErrorCodes.SETTINGS_ERROR:
+      case ErrorCodes.TASK_ERROR:
+      case ErrorCodes.JOB_ERROR:
+      case ErrorCodes.THUMBNAIL_GENERATION_ERROR:
+      case ErrorCodes.PHOTO_GENERATION_ERROR:
+      case ErrorCodes.SERVER_ERROR:
+        return Logger.error
+      default:
+        return Logger.warn
+    }
   }
 }
--- a/src/backend/Logger.ts
+++ b/src/backend/Logger.ts
@@ -11,8 +11,7 @@ if (forcedDebug === true) {
   );
 }
 
-// Match `console.log` inputs
-export type LoggerArgs = Parameters<Console["log"]>[0]
+export type LoggerArgs = (string | number | (() => string) | Record<any, unknown> | Error);
 export type LoggerFunction = (...args: LoggerArgs[]) => void;
 
 export interface ILogger {
--- a/src/backend/middlewares/RenderingMWs.ts
+++ b/src/backend/middlewares/RenderingMWs.ts
@@ -132,11 +132,14 @@ export class RenderingMWs {
   ): void {
     if (err instanceof ErrorDTO) {
       if (err.details) {
+        const logFn = Logger.logLevelForError(err.code)
+
+        logFn('Handled error:');
         LoggerRouter.log(logFn, req, res);
         // use separate rendering for detailsStr
         const d = err.detailsStr;
         delete err.detailsStr;
-        console.log(err);
+        logFn(err);
         err.detailsStr = d;
         delete err.details; // do not send back error object to the client side
 
--- a/src/backend/middlewares/RenderingMWs.ts
+++ b/src/backend/middlewares/RenderingMWs.ts
@@ -134,12 +134,11 @@ export class RenderingMWs {
       if (err.details) {
         const logFn = Logger.logLevelForError(err.code)
 
-        logFn('Handled error:');
-        LoggerRouter.log(logFn, req, res);
         // use separate rendering for detailsStr
         const d = err.detailsStr;
         delete err.detailsStr;
-        logFn(err);
+        logFn('Handled error:\n', err);
+        LoggerRouter.log(logFn, req, res);
         err.detailsStr = d;
         delete err.details; // do not send back error object to the client side
 
--- a/src/backend/middlewares/RenderingMWs.ts
+++ b/src/backend/middlewares/RenderingMWs.ts
@@ -138,13 +138,6 @@ export class RenderingMWs {
         const d = err.detailsStr;
         delete err.detailsStr;
         console.log(err);
-        if (err.detailsStr) {
-          try {
-            console.log('details:', JSON.stringify(err.detailsStr));
-          } catch (_) {
-            console.log(err.detailsStr);
-          }
-        }
         err.detailsStr = d;
         delete err.details; // do not send back error object to the client side
 
--- a/src/backend/middlewares/RenderingMWs.ts
+++ b/src/backend/middlewares/RenderingMWs.ts
@@ -132,13 +132,11 @@ export class RenderingMWs {
   ): void {
     if (err instanceof ErrorDTO) {
       if (err.details) {
-        const logFn = Logger.logLevelForError(err.code)
-
+        LoggerRouter.log(logFn, req, res);
         // use separate rendering for detailsStr
         const d = err.detailsStr;
         delete err.detailsStr;
-        logFn('Handled error:\n', err);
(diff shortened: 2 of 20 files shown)


src/backend/Logger.ts
```diff
@@ -11,7 +11,8 @@ if (forcedDebug === true) {
   );
 }
 
-export type LoggerArgs = (string | number | (() => string))
+// Match `console.log` inputs
+export type LoggerArgs = Parameters<Console["log"]>[0]
```
bpatrik, line 15: This is not exactly true. There was a recent PR that made it possible to use string returning functions as parameter, so the args do not equal with console.log.

src/backend/model/database/GalleryManager.ts
```diff
@@ -123,7 +123,7 @@ export class GalleryManager {
         );
         ObjectManagers.getInstance()
             .IndexingManager.indexDirectory(relativeDirectoryName)
-            .catch(console.error);
+            .catch(Logger.error);
```
bpatrik, line 126: Are you certain this would work the same way?


To my memories, console.error properly prints the stacktrace of an error. It also prints errors to stderr, not to stdout.  not sure if the 2nd is an issue, but the stacktrace is needed.

Loggger uses console.log only.

src/backend/model/extension/ExtensionConfigWrapper.ts
```diff
@@ -3,7 +3,7 @@ import {PrivateConfigClass} from '../../../common/config/private/PrivateConfigCl
 import {ConfigClassBuilder} from 'typeconfig/node';
 import {ExtensionConfigTemplateLoader} from './ExtensionConfigTemplateLoader';
 import {NotificationManager} from '../NotifocationManager';
-
+import {Logger} from '../../Logger';
```
bpatrik, line 6: This is very tricky here.  This class loads the config. Logger depends on config. Using logger in this class would end in circular dependency,  so it could not be used here.

src/backend/model/messenger/StdoutMessenger.ts
```diff
@@ -12,6 +13,6 @@ export class StdoutMessenger extends Messenger {
 
 
   protected async sendMedia(config: never, media: MediaDTOWithThPath[]) {
-    console.log(media.map(m => m.thumbnailPath));
```
bpatrik, line 15: This class is kind dummy/sample class. Its purpose is to print some data to the sdtout without any other text, so its not really logging here.

bpatrik (changes requested):

bpatrik:
Thank you very much for your contribution. 

I don't mind if login or other scraper errors are supressed . Moving /api /pgapi and updating nginx path helped a lot to reduce these attempts.

src/backend/Logger.ts
sarayourfriend: True, `console.log`'s args are all `any`, so it happens to work. I could change this either to explicitly `any`, or try to expand the types of the original `LoggerArgs` to accept things like the `ErrorDTO`. Adding `| any` to the existing type reduces it to just `any` in the end anyway.

The underlying implementation of the `Logger` functions remains the same, so passing a string-returning function will continue to work exactly as before... there's just no indication in the types that that's an _expected_ input :thinking: 

A type something like `string | number | (() => string) | object` might capture everything correctly and still indicate in the type that a string-returning function is expected. What do you think?

src/backend/model/database/GalleryManager.ts
sarayourfriend: You're right, there is a difference. I considered changing the `Logger.error` implementation to use stderr to match `console.error`'s behaviour, and make things more consistent with typical expectations for node console. In reading some of the related issues, however, I read [this comment of yours](https://github.com/bpatrik/pigallery2/issues/519#issuecomment-1344832911) where you said the app "only prints logs to stdout". I assumed this was the intention, but I should have clarified that in the PR description! I did not consider the difference in the stack trace printing, total oversight, thank you for catching it.

If it's not necessarily a strict intention to send everything to stdout, then changing the implementation of `Logger.error` to use `console.error` would keep the behaviour of places like this equivalent (other than adding the `Logger.error` prefix, but would introduce changes in the behaviour for existing cases of `Logger.error`.

What's your preference here? I can also remove this clean up from the PR, it's a bit outside the specific scope of changing the log-level of the relevant error types :thinking:

src/backend/model/messenger/StdoutMessenger.ts
sarayourfriend: I'll revert this change.

src/backend/model/extension/ExtensionConfigWrapper.ts
sarayourfriend: Oh, I didn't catch that, I guess I didn't test with an extension locally. I'll just revert this along with the other logging "clean ups" as outside the scope of the log-level changes.

sarayourfriend:
Thanks for the revievw @bpatrik. I mentioned in a couple of the review comments, but to clarify, I'll go ahead and make the following changes:

1. I'll set back `LoggerArgs` to the original type with `| object` added, to allow passing error objects for to leveral console.log's printing.
2. I'll undo the broader logging "clean up" and stick to just the log-level changes. That'll cut the scope and make this easier to review. Sorry for the noise there!

src/backend/Logger.ts
sarayourfriend: Ah, trying this locally, `| object` widens the type to essentially `any`. `() => 123` is accepted now, when previously it wouldn't have been. Incidentally, it does work, because there's nothing in the `Logger.log` function that expects the return type of the functions to be a string specifically, it just needs to be something `console.log` will accept as an argument (which is `any`).

A narrower type like `Record<any, unknown>` doesn't allow for the Error objects, but adding a further `| Error` keeps the stronger argument type than `any` while still opening up passing the error objects in `RenderingMWs`.

```ts
export type LoggerArgs = (string | number | (() => string) | Record<any, unknown> | Error);
const t: LoggerArgs = () => 123  // => not assignable
```

```ts
export type LoggerArgs = (string | number | (() => string) | object);
const t: LoggerArgs = () => 123 // => allowed, `LoggerArgs` is basically `any`
```

It feels a little fiddly to do it this way, considering `console.log` takes `any` anyway, but it's nice to have the types stay the same and reduce the potential scope of changes.

src/backend/model/database/GalleryManager.ts
sarayourfriend: Resolving the comment as I've reverted these changes in the PR.

src/backend/model/messenger/StdoutMessenger.ts
sarayourfriend: Reverted the change.

src/backend/Logger.ts
```diff
@@ -1,7 +1,7 @@
 import {Config} from '../common/config/private/Config';
 import {LogLevel} from '../common/config/private/PrivateConfig';
+import {ErrorCodes} from '../common/entities/Error';
 
-export type logFN = (...args: (string | number | (() => string))[]) => void;
```
sarayourfriend, line 4: This type duplicated the `LoggerFunction` type below, and was only used in one place, so I consolidated them.

src/backend/middlewares/RenderingMWs.ts
```diff
@@ -132,19 +132,12 @@ export class RenderingMWs {
...
         const d = err.detailsStr;
         delete err.detailsStr;
-        console.log(err);
-        if (err.detailsStr) {
-          try {
-            console.log('details:', JSON.stringify(err.detailsStr));
-          } catch (_) {
-            console.log(err.detailsStr);
-          }
-        }
```
sarayourfriend, line 147: This code is dead. `err.detailsStr` is deleted on 139, so the conditionl will always evaulate to false and the block won't run.

I'm not familiar with this code to interpret what the _intention_ was. If you know what it's meant to do @bpatrik I can update it to do that instead of removing the code.

bpatrik: To be honest I can't remember either. 

Removing it sounds good to me.

bpatrik (approved):

bpatrik:
Thank you! 

This change looks good to me.

Issue #865, opened 2024-03-29
Question: Metadata xmp.rating values


grasdk:
Question1: What are valid values of `xmp.rating`?

Background:
- [Adobe defines](https://developer.adobe.com/xmp/docs/XMPNamespaces/xmp/) the legal values as `{-1, 0, 1, 2, 3, 4, 5}` with `[1, 5]` as rating values, `-1` as a special "rejected" value and `0` meaning "unrated". Furthermore, adobe specifies that no rating is assumed to be a `0`.

The UI seems to direct the user towards values 0 to 5 only.

Question2: Assuming that pigallery adopts the range as specified by adobe: If a value outside the range `-1` to `5` is read, what is the expected behavior of pigallery: Default to `0` or to nearest valid value?
Example ideas for "nearest value":

- `-10` will map to `-1`
- `-100` will map to `-1`
- `10` will map to `5`
- `100` will map to `5`
- `2.1` will map to `2` (rounding)
- `2.4` will map to `2` (rounding)
- `2.5` will map to `3` (rounding)
- `2.9` will map to `3` (rounding)
- `3.1` will map to `3` (rounding)
- any non-numerical value will map to `0`
- no rating defined will map to `0`

bpatrik:
I think for now the app support rating of 0 to 5 and undefined (null in db). If null nothing is shown on the UI otherwise the given numbers of starts.

I'm only mediumly certain about this, need to double-check. (Null and 0 maybe handled the same way, which is probably a bug)


I don't mind if we change this, but then grouping, sorting and searching should be updated and kept consistent. ( Preferably in one PR).

In Adobe case what is the difference between rejected and unrated?

Or reverted is when you hit 'x' and the photo turns gray in Lightroom?
That sounds to me a Lightroom specific UI feature and I think pg2 should not follow it, unless it's a more widely adopted feature.

 
Answer 2:

That mapping sounds more or less good to me.

grasdk:
> I'm only mediumly certain about this, need to double-check. (Null and 0 maybe handled the same way, which is probably a bug)
There was a recent PR, where actual 0 values were treated like not present. So if a value 0 was read from actual metadata, the pigallery metadata would show nothing. The opposite of what the [spec](https://developer.adobe.com/xmp/docs/XMPNamespaces/xmp/) says: 
_If xmp:Rating is not present, a value of 0 should be assumed_

[...]

PR #868, opened 2024-03-30
Metadata mapping and consolidation

grasdk:
* Removed the SideCar Type to make reading sidecar data just as generic as reading the embedded exif data
* Moved metadata-mapping into separate function to prepare for reuse
* Most sidecar metadata is now read with the same function as embedded metadata (for photos). Fixed a bug with rating: 0 being ignored
* added new test for special chars and timezone
* merge incoming changes
* xmp rating fix
* refactored mapping of metadata into smaller functions
* consolidated GPS coordinate mapping
* simplification of sidecar date reading logic
* Moved date mapping code into one function - fixed testdata 
* CreationDateOffset is now mapped from Sidecar timestamps. Tests updated accordingly.
* Changed sidecar reading for video to shared reader and fixed tests accordingly (more data is read now)


grasdk:
For some in depth info, take a look at each commit. It shows the "journey" in smaller steps:
https://github.com/bpatrik/pigallery2/compare/master...grasdk:pigallery2:feature/metadata-sidecar-reading-cleanup

grasdk:
Hey @bpatrik 
So this was meant to make it easier to add future fixes (new tags to read metadata from, or even new metadata), in a way that it will both embedded data in photoes and sidecar data for both photos and videos.

In the process I also stumbled over xmp.rating, which I asked [a question](https://github.com/bpatrik/pigallery2/issues/865) about. I took the liberty of updating according to the adobe spec (adding -1 as legal value) in this pull request, but it can be easily undone, if you would rather not have that.
- if keeping it, should I bump the database version again to ensure that running installations get their databases rebuilt?

grasdk:
I found out I'm missing the -1 min value in `query-entry.search.gallery.component.html`. I will add this of course, unless you'd rather want the rating part reverted.

Extra question: Should I also add the "no rating" means rating 0 in pigallery, is the adobe specification suggests is the intention?

bpatrik:
Huge thanks for the refactoring! This an addition that also wanted to do myself, but I find it hard recently spend time on this project and also I'm getting lost in the extension configuration labyrinth. 

Would you mind separating the -1 change into a separate PR from the refactoring part?

For PRs I preffer having as small as possible as it's easier to review, but at least as big as merging it still makes production coherent (lot of people uses the edge build, so if it builds it should be consistent).

For the -1 handling I will answer in separate thread.

bpatrik:
> Hey @bpatrik 
> So this was meant to make it easier to add future fixes (new tags to read metadata from, or even new metadata), in a way that it will both embedded data in photoes and sidecar data for both photos and videos.
> 
> In the process I also stumbled over xmp.rating, which I asked [a question](https://github.com/bpatrik/pigallery2/issues/865) about. I took the liberty of updating according to the adobe spec (adding -1 as legal value) in this pull request, but it can be easily undone, if you would rather not have that.
> - if keeping it, should I bump the database version again to ensure that running installations get their databases rebuilt?

Yes if reindexing is needed (which in this case, would), that should be bumped

grasdk:
Ok, xmp.rating now works as before, 0 and below maps to "undefined" (also for embedded metadata). Values larger than 5 maps to 5. In case of decimal ratings, I added rounding as well.
This should be in line with your comment, and a suitable part of the refactoring :)

>Based on this I think rating **should work as follows**:
>* Undefined,  null and 0 are equivalent  and  metadat.rating should not be set.
>   * searching for 0 rating should list all of these photos 
>* Valid ratings are 1-5. UI only shows rating if its between 1-5.
>* -1 mapped to 0

_Originally posted by @bpatrik in https://github.com/bpatrik/pigallery2/issues/865#issuecomment-2028452710_

src/backend/model/database/enitites/MediaEntity.ts
```diff
@@ -132,7 +132,7 @@ export class MediaMetadataEntity implements MediaMetadata {
   @Column(() => PositionMetaDataEntity)
   positionData: PositionMetaDataEntity;
 
-  @Column('tinyint', {unsigned: true})
+  @Column('tinyint', {unsigned: false})
```
bpatrik, line 135: This should stay unsigned.

bpatrik (changes requested):

bpatrik:
Great, thank you very much! :)

grasdk:
Thank you too :)

Another step for [this old enhancement request](https://github.com/bpatrik/pigallery2/issues/156), since the refactor adds more support for video metadata, due to the consolidation of metadata handling. :)

Issue #14, opened 2018-09-06
Resource management


angelmanst:
Hi,

Whenever I go into a folder that has more than 20 photos (I have some would 200+), my raspberry pi 3+ slows down a lot (stops responding even to ssh sessions). I think I've narrow the problem down to the use of swap during thumbnails creation of PiGallery2.

Would it be possible to limit the number of thumbnails to be created in parallel so that memory usage wouldn't be so high that it would need to swap?

Regards,
Marco Angélico

tomikais:
Hi @angelmanst, Hi @bpatrik,

so I think I noticed the same behavior. Could you scan for new images after a period of time and then create the thumbnails?

Regards, Thomas

bpatrik:
Hi all,

Thats interesting, I don't experience anything like that on a RPi 3. 
I'm using `sharp` for rendering thumbnails. 

Anyways, I'll add a limit some point. 
In the meantime if it really annoys you, you can do the following:

1) Disable threading in the config.

2) a local hotfix:
at [ThumbnailGeneratorMWs.ts:31](https://github.com/bpatrik/pigallery2/blob/151a3782acb6bc00fc49ff97e82352dedfacb538/backend/middlewares/thumbnail/ThumbnailGeneratorMWs.ts#L31) 
change this:
` Config.Client.concurrentThumbnailGenerations = Math.max(1, os.cpus().length - 1);`
to this:
` Config.Client.concurrentThumbnailGenerations = 1;`
so it will only create thumbnails only on 1 thread.

(Warning you might find the same property in the config file, but modifying there wont take effect).
Of course it requires rebuild if you are doing it in the `*.ts` files. (You can also change it in the compiled `*.js` file, so you don't have to rebuild)
Patrik

bpatrik:
With the latest commit [48aaf54](https://github.com/bpatrik/pigallery2/commit/48aaf54c7d3185eba0602060869b503c117aeb65), I've created an 'Index with Thumbnail' button that scans all directories, indexes the photos/videos and creates their (smallest) Thumbnails too.

bpatrik:
I'm closing this due to inactivity

PR #870, opened 2024-03-31
Feature/caption and title (#14)

grasdk:
- reading caption and title from several tags with several tests
- added minor bugfix for date conversion to MS
- optimization: break sidecar loop, when first sidecar is found


grasdk:
Ok. Once again I'm struck by dates. The error has nothing to do with this PR.

The problem lies in these lines:
https://github.com/bpatrik/pigallery2/blob/master/src/backend/model/database/SearchManager.ts#L769-L790

When running the tests today, the 31st of march, the subtraction of months behaves weirdly. Javascript data handling is notoriously bad.

Take a look here, how subtracting a month from the 31st of march (or 1st of april) suddenly becomes the 3rd of march:

```
>to = new Date();
Date Sun Mar 31 2024 17:39:12 GMT+0200 (Central European Summer Time)

>to.setHours(0, 0, 0, 0);
1711839600000
>to.setUTCDate(to.getUTCDate() + 1);
1711926000000
>to
Date Mon Apr 01 2024 01:00:00 GMT+0200 (Central European Summer Time)

>to.setUTCMonth(to.getUTCMonth() - 1);
1709420400000
>to
Date Sun Mar 03 2024 00:00:00 GMT+0100 (Central European Standard Time)

>to.setUTCMonth(to.getUTCMonth() + 1);
1712098800000
>to
Date Wed Apr 03 2024 01:00:00 GMT+0200 (Central European Summer Time)
```

I will try and fix it nevertheless.

bpatrik:
Ahh this search is annoying.


I would just create a PR for the original feature.  The code looks good to me so far. If metadata  test runs, I will merge it.

Then we can have a separate discussion or at least a bug about the date search.

grasdk:
Ok. I think my solution to the search is more intuitive than the default js one.

Basically it is:
If (date+x months)'s day of month does not exist, change to the largest day of month.
- So 31st of april, june, september and november are converted to 30th.
- 30th and 31st of february are converted to 29th on leap years
- 29th, 30th and 31st of February are converted to 28th of non-leap years

So 31st of march one month ago and 3 days back means, 27th, 28th and 29th of February this year.
Before my change it was 1st, 2nd and 3rd of march this year.

Of course, I can remove this easily from the PR if you would like that. Can do it tonight. :-)

grasdk:
I reverted not only the tests but also the other search-mechanism described above.

I will resubmit in another PR, so you can try it out separately :)

bpatrik:
Thank you!

Issue #872, opened 2024-04-02
Images Missing "Date/Time Original" All Arbitrarily Assigned the Same Date


grainsoflight:
I have 2000+ Images in my collection, many of which have had the "Date/TIme Original" Metadata altered in lightroom with dates ranging from 1-1-1900 to present. Everything that was altered shows up correctly. However anything that I did not make any changes to from the metadata that was written when I scanned the image gets dumped into 1-1-1970 on the timeline. When opening the information tab, there is no date info to suggest it should be in the 1970's, and when read in lightroom the image dates will be from sometime in the last 3 years when they were scanned.

**Environment (please complete the following information):**
 - OS: Unraid
 - Browser: Chrome

**Used app version**:

- docker:edge-alpine

grainsoflight:
@grasdk do you think this might be related to my other issue?

grasdk:
Hey.

Yes it can be related in the manner that your unaltered pictures store their date in a field not read by pigallery. There are an insane amount of date fields.

Can you provide a sample, or if not, then maybe show the output of running exiftool on one of the pictures?

For example this command could provide useful info

`exiftool -G0:1 -*date* -*time* -*offset* mypicture.png`

grainsoflight:
@grasdk for instance, the previous image I submitted had it's date changed to some date in 1900 in lightroom, however this image never had it's capture date changed and instead should only include the date it was scanned into the computer
![family-1117](https://github.com/bpatrik/pigallery2/assets/29415011/9b6ea306-163d-4468-b3ef-83d4cc29ed83)

grasdk:
thanks. It's a bug. The timestamp is actually stored in a field that is read already, but there is a bug in the code where the formatted timestamp is converted into a MS value.

It can easily be fixed, so I'll make sure to make a PR for that very soon.

grainsoflight:
@grasdk Thanks, just to be clear, "Date Created" will still take precedence over "Create Date," correct?

grasdk:
I think so, but I will try and make it clear with the PR. The tags have different names than the ones you mention, which are shown by your OS or image viewer.

In your photo, the tag `xmp:CreateDate` will be used. There are others that have higher priority, but they are not set in your photo.

grainsoflight:
[...]

PR #874, opened 2024-04-03
Feature/timestamp fix

grasdk:
Handling of timestamp with negative offset fix


grasdk:
Hey. That fix was merged 5 days ago. Which image are you using? It seems like only edge-alpine has been built since:
https://hub.docker.com/r/bpatrik/pigallery2/tags/

bpatrik:
> Hey. That fix was merged 5 days ago. Which image are you using? It seems like only edge-alpine has been built since: https://hub.docker.com/r/bpatrik/pigallery2/tags/

I'm not sure what is the question for this answer :)

grasdk:
Woops, it was a question from the issue. I must have posted here by mistake. I wondered why my reply was gone :)

Here is the correct thread, so please ignore the comment above :)
https://github.com/bpatrik/pigallery2/issues/872#issuecomment-2046531114

PR #875, opened 2024-04-04
Feature/month addition

grasdk:
What does 1 month back mean, when it's 31st of march 2020 (which was a leap year)?

Does it mean 31st of february, which doesn't exist, so 31st of february becomes 29th of february + 2 days, i.e. 2nd of march?
- That's how Javascript and therefore Typescript does it, when you subtract one from month, using the Date.setMonth function.

I think that when you subtract a month, you HAVE to end up in the previous month. And the consequence is that 31st of March 2020 minus 1 month becomes 29th of february (and even the 28th on non-leap years).

- 30th and 31st of march minus 1 month becomes 29th of february on leap years
- 29th, 30th and 31st of march minus 1 month becomes 28th of february on non-leap years
- 31st of may, july, october and december minus 1 month becomes 30th of april, june, september and november respectively.

This PR alters the behavior or x months subtration in this manner. :)


bpatrik:
aggh that is rough. 
Thanks for finding and fixing it!

Issue #877, opened 2024-04-05
edge-alpine DB fails with an HEIC image for media_entity.metadataSizeWidth constraint


malavolti:
----------------------------------------------------------------------------------

**Describe the bug**

If I try to load this image, I receive the following error on SQLLite DB and no images are loaded on my piGallery2.

**Photo/video (optional) that causes the bug**

https://we.tl/t-pEdQHVDG4W

**Server logs (optional)**

```
QueryFailedError: SqliteError: NOT NULL constraint failed: media_entity.metadataSizeWidth
    at BetterSqlite3QueryRunner.query (/app/node_modules/typeorm/driver/better-sqlite3/BetterSqlite3QueryRunner.js:103:19)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async InsertQueryBuilder.execute (/app/node_modules/typeorm/query-builder/InsertQueryBuilder.js:106:33)
    at async SubjectExecutor.executeInsertOperations (/app/node_modules/typeorm/persistence/SubjectExecutor.js:260:42)
    at async SubjectExecutor.execute (/app/node_modules/typeorm/persistence/SubjectExecutor.js:92:9)
    at async EntityPersistExecutor.execute (/app/node_modules/typeorm/persistence/EntityPersistExecutor.js:148:21)
    at async IndexingManager.saveChunk (/app/src/backend/model/database/IndexingManager.js:404:32)
    at async IndexingManager.saveMedia (/app/src/backend/model/database/IndexingManager.js:329:9)
    at async IndexingManager.saveToDB (/app/src/backend/model/database/IndexingManager.js:95:13)
    at async IndexingManager.queueForSave (/app/src/backend/model/database/IndexingManager.js:128:17) {
[...]

PR #879, opened 2024-04-05
updated image-size dependency to read dimensions from heic file

grasdk:
Added extra not-null assurance for image width and height

updated image-size library (still has some bugs, though).


grasdk:
Fix for https://github.com/bpatrik/pigallery2/issues/877

It's not a complete fix though, but it should stop the SQL error.

However, the latest version of the image-size module does not correctly report the size of the test heic-image that was added. I adapted the test to work with what image-size returned.

src/backend/model/fileaccess/MetadataLoader.ts
```diff
@@ -206,6 +206,13 @@ export class MetadataLoader {
       } catch (e) {
         //in case of failure, set dimensions to 0 so they may be read via tags
         metadata.size = { width: 0, height: 0 };
+      } finally {
+        if (metadata.size.width == undefined || metadata.size.width == null) {
```
bpatrik, line 210: Maybe isNaN?

bpatrik (changes requested):
Some not important feedback. Feel free to discard

src/backend/model/fileaccess/MetadataLoader.ts
grasdk: isNaN(null) is actually false, but replacing the == undefined with isNaN is a good idea, if the value should somehow ever be contaminated with values that are not numbers. :)

I will change this, but since I expect a bit of conflict with my other PR, I might as well fix these conflicts and this comment in one go :)

Issue #847, opened 2024-03-07
Extension configuration is broken


mblythe86:
**Describe the bug**

The values from `extension.config.getConfig()` no longer reflect what is present in `config.json`, nor what is shown on the admin page.

**Screenshots (optional)**

This is present in my config.json:

```
    "Extensions": {
        "enabled": true,
        "//[folder]": "Folder where the app stores all extensions. Individual extensions live in their own sub-folders.",
        "folder": "/app/data/config/extensions",
        "extensions": [
            {
                "enabled": true,
                "//[path]": "Folder where the app stores all extensions. Individual extensions live in their own sub-folders.",
                "path": "dk-extension",
                "configs": {
                    "//[digikamShowCollection]": "DigiKam Directory Category",
                    "digikamShowCollection": "Public",
                    "//[digikamDbType]": "DigiKam Database Type (MySQL or SQLite)",
                    "digikamDbType": "MySQL",
                    "//[digikamSqliteDb]": "DigiKam SQLite DB filename",
                    "digikamSqliteDb": "/app/data/digikam/digikam.db",
                    "//[digikamMysqlHost]": "DigiKam MySQL DB hostname",
                    "digikamMysqlHost": "192.168.42.3",
                    "//[digikamMysqlPort]": "DigiKam MySQL DB port",
                    "digikamMysqlPort": 3306,
                    "//[digikamMysqlDb]": "DigiKam MySQL DB name",
                    "digikamMysqlDb": "digikam",
                    "//[digikamMysqlUser]": "DigiKam MySQL DB username",
                    "digikamMysqlUser": "dk_ro",
                    "//[digikamMysqlPassword]": "DigiKam MySQL DB password",
                    "digikamMysqlPassword": "redacted"
                }
            }
        ],
        "//[cleanUpUnusedTables]": "Automatically removes all tables from the DB that are not used anymore.",
        "cleanUpUnusedTables": true
    },

```

This is what the admin page shows:

![image](https://github.com/bpatrik/pigallery2/assets/751675/2cf84239-8144-45f1-9ea2-d4a04ae8d1be)

**Server logs (optional)**

When I have the extension dump the `extension.config.getConfig()` object, the state has the default values, not my real values:

```
pigallery2_dev  |   __state: {
pigallery2_dev  |     digikamShowCollection: {
pigallery2_dev  |       description: 'DigiKam Directory Category',
pigallery2_dev  |       type: 'string',
pigallery2_dev  |       value: 'Public',
[...]

PR #884, opened 2024-04-12
Fixes for extension config loading (issue #847)

mblythe86:
* Actually add/remove extensions from `Config.Extensions.extensions`
* Fix the file path where `ExtensionConfigTemplateLoader` looks for extensions
* Prevent the first `loadSync()` call from creating a `config.json` file
  Works around part of the issues caused by bpatrik/typeconfig#27

This seems to fix the problems of generating a fresh config with extension settings, saving extension config changes to the file, and loading extension settings from a config file that already includes them.

As noted in #847, this isn't a complete fix.  There's still the problem of adding an extension to an existing install.  I don't think that can be fixed in pigallery2.  I think that needs to be fixed in typeconfig.

### DIFF
--- a/src/backend/model/extension/ExtensionManager.ts
+++ b/src/backend/model/extension/ExtensionManager.ts
@@ -12,6 +12,7 @@ import {SQLConnection} from '../database/SQLConnection';
 import {ExtensionObject} from './ExtensionObject';
 import {ExtensionDecoratorObject} from './ExtensionDecorator';
 import * as util from 'util';
+import {ServerExtensionsEntryConfig} from '../../../common/config/private/subconfigs/ServerExtensionsConfig';
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const exec = util.promisify(require('child_process').exec);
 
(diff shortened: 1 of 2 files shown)


src/common/config/private/Config.ts
```diff
@@ -6,8 +6,12 @@ import * as path from 'path';
 
 const pre = ConfigClassBuilder.attachPrivateInterface(new PrivateConfigClass());
 try {
+  //NOTE: can possibly remove this saveIfNotExist hack if typeconfig issue #27 is fixed
+  const origSaveIfNotExist = (pre.__options as any).saveIfNotExist;
```
bpatrik, line 10: I'm trying to do something very similar here: https://github.com/bpatrik/pigallery2/commit/3380fef63ae3366be357202f550789e455415264

src/backend/model/extension/ExtensionManager.ts
```diff
@@ -79,6 +80,14 @@ export class ExtensionManager implements IObjectManager {
       );
     extList.sort();
 
+    // delete not existing extensions
```
bpatrik, line 83: i wanted to avoid this. (i.e.: I removed this laoding here on porpuse.)

I would like to have a single place that is responsible loading extension config (even if they do not exists, then its an empty config). 
That is why I moved all code to the extension config loader.

src/common/config/private/Config.ts
bpatrik: Ah I was wrong that commit was fixing a different issue. But I eventually came to the same conclusion with this issue: https://github.com/bpatrik/pigallery2/commit/055862f2753f621fb521b8c9b68731cbc709dc2e

bpatrik:
I will reject this as fixes for the same issue should be already part of the master

src/common/config/private/Config.ts
mblythe86: Agreed, it looks like the `{preventSaving: true}` should accomplish the same thing.

src/backend/model/extension/ExtensionManager.ts
mblythe86: I'll take a closer look at the extension config loader.  I didn't see anywhere else that `Config.Extensions.extensions.push(new ServerExtensionsEntryConfig(ep)));` happened, so the extension configs never actually got loaded into `Config`.

bpatrik:
I will reject it as I think the issue that this PR try to solve is already fixed.
Sorry and thank you for the effort. I appreciate it!

PR #891, opened 2024-04-18
Feature/clear date time tag priority

grasdk:
Rewrite of metadata date handling. Since there are over 150 different date/timestamp tags, some new requirements will probably pop up over time. 

In the effort of keeping MetadataLoader relatively simple, this PR adds a simple list of tags (in the notation of the exifr library) in a stand-alone file

`src/backend/model/fileaccess/MetadataCreationDate.ts`

The list is a prioritized list of tags that MetadataLoader is using to lookup `creationDate` and `creationDateOffset`. The first one found is used. If it does not contain offset, offset will be searched for in other tags under the conditions that these timestamps are close enough to the first.

The new solution allows any contributer to add more tags to the list or change the priority of the tags (if there is a good reason). The priority currently used, puts standardised exif first and has photoshop shortly after and leaves various modification dates at the bottom, since modification can be done long time after creation.

As before side car metadata are loaded last, which means that sidecar data will "overwrite" embedded data in Pigallery2's metadata (not in the real photos).

- This new solution identified a date calculation error in the test for `flatxmp.xmp`, which has now been fixed.

Also added some more test data


src/backend/model/fileaccess/MetadataCreationDate.ts
```diff
@@ -0,0 +1,29 @@
...
+//tagname1 is typically a full date time, but in some cases tagname1 and tagname2 together make up a full timestamp. The type makes it easier to code correct concatenation of tag1 and tag2
+
+//Interesting exiftool forums posts about some of these tags:
+//https://exiftool.org/forum/index.php?topic=13170.msg71174#msg71174 - about the meaning of exif.DateTimeOriginal, exif.CreateDate/exif.DateTimeDigitized and exif.ModifyDate
+//https://exiftool.org/forum/index.php?topic=15555.msg83536#msg83536
+
+//This is the PRIORITIZED LIST of tags which Pigallery2 uses to determine the date of creation of pictures.
+//The list is used for embedded picture metadata and xmp-sidecar files for both pictures and vidoes.
+
+export const DateTags: [string, string, string][] = [
```
bpatrik, line 11: Can you add documentation or better typing about the DateTag type here.
I would just probably declare a type timesptamp = string, typeoffset= string, etc..

bpatrik:
Thank you! This is a very nice addition!

src/backend/model/fileaccess/MetadataCreationDate.ts
grasdk: Good idea. I made a commit with something that looks a bit like that. Can be changed of course. Let me know ;)

bpatrik: thank you, its way easier to understand.

PR #897, opened 2024-04-28
Fix code that resolves the extension directory path

mblythe86:
I think this is the last fix needed for bpatrik/pigallery2#847.

### DIFF
--- a/src/common/config/private/Config.ts
+++ b/src/common/config/private/Config.ts
@@ -10,6 +10,6 @@ try {
   pre.loadSync({preventSaving: true});
 } catch (e) { /* empty */
 }
-ExtensionConfigTemplateLoader.Instance.init(path.join(__dirname, '/../../../../', pre.Extensions.folder));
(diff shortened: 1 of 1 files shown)


src/common/config/private/Config.ts
```diff
@@ -10,6 +10,6 @@ try {
   pre.loadSync({preventSaving: true});
 } catch (e) { /* empty */
 }
-ExtensionConfigTemplateLoader.Instance.init(path.join(__dirname, '/../../../../', pre.Extensions.folder));
+ExtensionConfigTemplateLoader.Instance.init(pre.Extensions.folder);
```
bpatrik, line 13: I'm not sure if that works in all cases. Let me take a look.

bpatrik: yeah so the docker used absolute path, while my test setup used a relative one. 
Your setup fixes it for absolute path (eg the docker build), but breaks it for absolute.

My last commit should fix it for both.

mblythe86:
Fixed by a different commit, no longer necessary.

Issue #883, opened 2024-04-10
Increase sharingKey complexity


VirtualGit:
**Is your feature request related to a problem? Please describe.**
Sharing an album is possible without password thanks to a sharing key. That's exactly my need. But I find the sharing key complexity absolutely too much weak. It is currently composed of 8 characters including [0-9a-f]. From my point of view this is too weak.

**Describe the solution you'd like**
I'm interested in having a much more secure sharing key. At least 32 characters, maybe 64, instead of 8. Even better, characters could be taken from [0-9A-Za-z], which increase number of combinations. Maybe the sharing key length could be a setting ?

**Describe alternatives you've considered** (optional)
No alternative.

I can work on it and push a MR if you want. I'm not familiar with open-source collaboration, but there's a first time for everything :) If so, tell me the direction you'd prefer.

VirtualGit:
FYI, I've done it here :
https://github.com/VirtualGit/pigallery2/commit/a4326a0f2e7798b798ae27b0fb7bf86d08b688ef
I need to do more tests, but it's looks good.
* Still missing translation entries (not sure how to update xlf files)
* Long keys are not wrapped in the share list (config page) ; I should maybe enable text ellipsis or something like that. Will think about it soon.

bpatrik:
Hi,
That commit looks good to me. Happy to merge it if you send a PR.

VirtualGit:
Is this ok for you :
(I'm not CSS expert !)
https://github.com/VirtualGit/pigallery2/commit/edf03fe2b1c5d8c5bc5a080e0874c3c679cf20e6

That leads to :
![screenshot_20240502_220320](https://github.com/bpatrik/pigallery2/assets/14853220/e129a2b8-f2eb-4868-b297-ada97f94cee4)
and :
![screenshot_20240502_220416](https://github.com/bpatrik/pigallery2/assets/14853220/fb7e0b69-1f38-4915-9614-29287cd52c21)

bpatrik:
Yeah, it looks good to me!

-- Sorry for being brief, sent from my phone.

bpatrik:
should be fixed now.

VirtualGit:
Thank you for the merge, and sorry for not having time to submit a PR. Regards.

PR #901, opened 2024-05-11
Sharing key increase complexity

VirtualGit:
Add a sharing key complexity setting to generate more complex keys.
Keys too long will be displayed with an ellipsis.
(refs #883)
That's my first github PR. I updated my branch twice so there are merge commits. Tell me if this is ok or if you want I make a new more clean branch.
Thanks


src/common/config/public/ClientConfig.ts
```diff
@@ -304,6 +304,16 @@ export class ClientSharingConfig {
...
   })
   passwordRequired: boolean = false;
+  @ConfigProperty({
+    type: 'unsignedInt', min: 8, max: 64,
+    tags:
+      {
+        name: $localize`Sharing key length`,
+        priority: ConfigPriority.underTheHood
+      },
+    description: $localize`The longer the keys are, the more secure the share links will be.`,
```
bpatrik, line 314: I would rather say something like: that longer the key, harder is it to guess it. The current phrasing implies better underlying security with longer keys.

I would also add that using lenght of 8 takes 70000 years to try all permutation (assuming the server can serve a request in 10s), which should be complex enough.

Also i would add a note that changing this number won't invalidate existing sharing

bpatrik (changes requested):

bpatrik:
Nice! Thank you! I made a comment on the config description. Otherwise it looks good to me.

src/common/config/public/ClientConfig.ts
VirtualGit: Ok, I changed to "The longer the keys are, the harder they are to guess.".
Do you want I add "Changing this number won't invalidate existing sharing" in the setting description ?

VirtualGit: Also, should I do a new PR with these changes, or can I update this one ?

bpatrik: Yes please add that text too. 
Yes you should be able merge the changes here too.

VirtualGit:
Sorry for the close, I had some troubles with PR. I'm just submitting a new one.

Issue #892, opened 2024-04-19
Search by date filters by UTC timestamp


mcdamo:
Search example: `after:2024-04-18 before:2024-04-19`

Expected result:
  * any media where timestamp was taken on '2024-04-18', regardless of timezones

Actual result:
  * media without a timezone offset shows as expected, because pigallery has assumed it to be in UTC.
  * media that has a timezone offset in metadata is excluded if it falls outside of the converted UTC timestamp of '2024-04-18`

mcdamo:
This is somewhat related to #871

bpatrik:
Yeah. This needs clean up.
We should fix this before the next release. (which I have no idea when will happen, but today I dont think it will before sept.)

@grasdk FYI. If you feel that you have the time for this. :)

grasdk:
@mcdamo I would say this bug is resolved in the latest edge version, if you leave the timestamp offset setting on "ignore". See https://github.com/bpatrik/pigallery2/pull/902

mcdamo:
Thanks @grasdk, however with "ignore" setting on, search by date now appears to exclude all images without timezones.

For example search `after:2024-04-13 and before:2024-04-15` is missing the `4MinsAroundTheWorld/20240413-203800-Unknown.jpg` image in search results.

grasdk:
@mcdamo 
Good spot. `null * 60000` is still `null` in SQLite (and others I guess), so had to add the coalesce-function to the "ignore" queries: `(coalesce(creationDateOffset,0) * 60000)`

I made a PR (https://github.com/bpatrik/pigallery2/pull/913), but the test is failing. Looks like a container error though, hopefully not caused by my code. You're welcome to pull and build my branch and see if it works for you.

grasdk:
@mcdamo my fix was merged last night and it appears that it was released unto the latest edge-version. Care to take it for a spin to see if you agree that it's fixed? :)

https://hub.docker.com/r/bpatrik/pigallery2/tags

mcdamo:
Thanks @grasdk, I tested the latest code and the search works with your fix.

Issue #871, opened 2024-04-01
Sort by date does not does not correctly handle timezones


mcdamo:
Sort by date does not does not correctly handle timezones.

In the case that a directory or search contains media with and without timezone data they will not display in the expected order.

Cameras generally store the _Local Time_ that the media was taken but modern cameras or smartphones can now record the Timezone offset as well.

The problem arising with sorting in Pigallery appears because all media is converted to UTC during indexing.

For example two photos taken at the same time, midday in GMT+09, one on a camera and the other on a smartphone
  1. on a camera **without timezone data** will be indexed as `12:00` with `0` offset
  2. on a smartphone **with timezone** will be indexed as `03:00` with `+09:00` offset. 

This means that the first photo will be sorted as if it were 9 hours later than the second within Pigallery.

Here in the code the sorting services uses the base UTC time without referencing the offset:
https://github.com/bpatrik/pigallery2/blob/6c53aca7e7d5b65d257c80737647d57854592bbe/src/frontend/app/ui/gallery/navigator/sorting.service.ts#L153

It may be more coherent to sort all media by the _Local Time_ instead of UTC.

bpatrik:
Thank you for the report.

@grasdk do you have a quick insight to this issue?

grasdk:
@bpatrik: I think @mcdamo already found the problem (or part of it).

It appears that I missed `sorting.service.ts`, when adding comments about searching and filetering should take offset into account, when I introduced offset to the metadata. 

I wrote some [//TODOs in the searchmanager](https://github.com/bpatrik/pigallery2/pull/829/files#diff-f85ded2f38ddf78f5cb870136c6928fe28a9532edf5ceaaee365c8120ad8ec29R566-R571) and in the [filter.service.ts](https://github.com/bpatrik/pigallery2/pull/829/files#diff-72075ba83c7886c7221fdcbdcd5912a9d95926d5c423f3de85aeb5f7cf93c333R208), but not in `sorting.service.ts`.

Basically, one can do something like

`return (a.metadata.creationDate + Utils.getOffsetMinutes(a.metadata.creationDateOffset)*60000) - (b.metadata.creationDate + Utils.getOffsetMinutes(b.metadata.creationDateOffset)*60000)`

in the line that @mcdamo writes.

[...]

PR #902, opened 2024-05-11
Bugfix/offset or ignore

grasdk:
New boolean option "Ignore timstamp offsets", default value true.

When enabled, searching, sorting, grouping and filtering uses the picture metadata's local time instead of the global time.

So two pictures taken at e.g. 11.00.00 AM 12 hours apart on opposite sides of the globe, will show up, when searching for dates that include this time.

Should solve: https://github.com/bpatrik/pigallery2/issues/871 and https://github.com/bpatrik/pigallery2/issues/892


grasdk:
@bpatrik, @mcdamo 

Added a Config setting "ignore timezone offset". When set, timestamps are handled as if they are local when searching and sorting. I also changed the handling in filtering, but did not see any effect on the drop-down filter, but I thought the code change was for the better anyway. New issue about the filter should probably be created, unless it's just me who don't understand how to use it ;)

bpatrik:
> @bpatrik Commenting on our talk. The changes here basically amount to
> 
>     * Adding a config parameter (`src/common/config/public/ClientConfig.ts`)
> 
>     * Adding an if-sentence to all database queries, in order to adjust the where clause (`src/backend/model/database)/SearchManager.ts`
> 
>     * Changing direct use of metadata.creationDate to a function call that adjusts whether or not to incorporate timestamp (all other files using `metadata.creationDate`)
> 
>     * * including `src/frontend/app/ui/gallery/filter/filter.service.ts`, where I don't get the expected effect on my demo-setup.


Can you describe what do you expect and what do you get? Do you maybe have a screenshot?

grasdk:
> Can you describe what do you expect and what do you get? Do you maybe have a screenshot?
With the "Ignore" enabled:
When the demo folder `4MinsAroundTheWorld` is opened, I would expect the filter to have min date the 13th and max date the 14th. Sliding the slider would remove pictures according to their local date.

With the "Ignor" disabled:
Same functionality but with 13th being both max and min date (but the slider still works for hours and minutes).
...
When wanting to provide screenshots, I did a total rebuild, and now the filter also works! :) So you can ignore my "bug" report og missing fix for this PR. :) All seems to work as expected.
> update: You will probably need to update the UI too (html file) to properly handle the timezone. As now the filter timezone number changes based on the config value.
I didn't have any issue here, except having to force a reload after switching the config toggle.

Issue #894, opened 2024-04-24
PiGallery2 stopped working - health issue - docker update ?


niawag:
Hi, I'm using PiGallery2 with docker (latest-alpine) on a x86_64 plateform with Ubuntu 22.04. I updated my docker version to version 26.1.0, build 9714adc (latest deb available) and compose to v2.26.1 and the update seems to have broken PiGallery2.
Here is my compose section:
```
pigallery2:
    image: bpatrik/pigallery2:latest-alpine
    container_name: pigallery2
    environment:
      - NODE_ENV=production 
      - TZ=Europe/Paris
      - PUID=1026
     - PGID=100
    networks:
      - t2_proxy
    security_opt:
      - no-new-privileges:true
    restart: always
    volumes:
      - "./pigallery2/config:/app/data/config"
      - "./pigallery2/db-data:/app/data/db"
      - "/mnt/photo:/app/data/images:ro"
      - "./pigallery2/tmp:/app/data/tmp"
    labels:
      - "traefik.enable=true"
      - "traefik.http.routers.pigall-rtr.entrypoints=https"
      - "traefik.http.routers.pigall-rtr.rule=Host(`pigall.mydomain.wtf`)"
      - "traefik.http.routers.pigall-rtr.tls=true"
      - "traefik.http.routers.pigall-rtr.service=pigall-svc"
      - "traefik.http.services.pigall-svc.loadbalancer.server.port=80"
      - "traefik.http.routers.pigall-rtr.middlewares=localip"
```
[The debug logs seems OK](https://pastebin.com/3iCqTnqL)

I tried with edge-alpine and lastest images, same issue. 
Is there someone else affected by this bug ? Is there anything I can do to help fixing that bug ?

niawag:
After some testing, it seems the latest docker version is responsible. I downgraded the following packages to these versions:
docker-ce=5:25.0.5-1~ubuntu.22.04~jammy
docker-ce-cli=5:25.0.5-1~ubuntu.22.04~jammy
docker-compose-plugin=2.25.0-1~ubuntu.22.04~jammy
(using "sudo apt install docker-ce=5:25.0.5-1~ubuntu.22.04~jammy" and so on)
And PiGallery is up and running. (tested with docker 26.0.2-1, not working either)

I have a lot of differents containers running on that server, all the others were running fine on the latest docker version so it must be something specific to PiGallery

bpatrik:
yeah those logs does not tell much. 
[...]

PR #920, opened 2024-06-30
Update Dockerfile.build - healthcheck changed from localhost to 127.0.0.1

niawag:
Should fix a bug whith IPv6 and healthcheck as shown in this issue: https://github.com/bpatrik/pigallery2/issues/894


bpatrik:
Thank you!

PR #923, opened 2024-07-27
Fix a few typos.

jose1711:


bpatrik:
Thank you for your contribution!

Can you please also update the English text in the original source file? (Mostly just html)

The English translation file is different from the rest as that is auto generated from the source files.

jose1711:
> Thank you for your contribution!
> 
> Can you please also update the English text in the original source file? (Mostly just html)
> 
> The English translation file is different from the rest as that is auto generated from the source files.

Actually, I am working on a Slovak translation and would like to avoid conflicts. Will first finish it, raise a PR and then come back to this. Thank you for your understanding.

jose1711:
Closing in favor of https://github.com/bpatrik/pigallery2/pull/931.

Issue #777, opened 2023-11-24
Stop using `VOLUME` in the Dockerfile


nk9:
**Describe the bug**

Dockerfiles have a `VOLUME` directive, but it's quite problematic and has no benefits over describing the volumes at runtime using `-v` or `docker compose`.

- Here's [an article](https://boxboat.com/2017/01/23/volumes-and-dockerfiles-dont-mix/) that goes over the issues. "So go ahead, use volumes, they are a great feature of Docker. Just don't define them inside your Dockerfile."
- And [a StackOverflow answer](https://stackoverflow.com/a/62068396/1749551) which also highlights these same problems

The recommended way to use piGallery2 with Docker is to use the `docker-compose.yml` file, which declares the volumes itself. Given that, it may be possible to just remove the `VOLUME` directive entirely.

bpatrik:
Hi,

Thank you for the notice!
Can send a PR fixing the docker build files and the compose files that
removes volumes from the build files, but keeps the current compose
behavior? (I mean keeps exposing config and images folder)

-- Sorry for being brief, sent from my phone.

nk9:
I lost track of this. Thanks for jumping in and fixing it, @grasdk!

PR #948, opened 2024-10-25
Feature/docker update

grasdk:
Resolves [issue 777](https://github.com/bpatrik/pigallery2/issues/777)

Also adds linting checks to dockerfiles and resolves linting errors. Linting errors about versioning packages are ignored (DL3008 and DL3018)


bpatrik:
Thanks. I did not test how removing the `Volume` tags would change the behavior, but I will just take your word that it will just keep working.

jdavidpeter:
To me it seems that the change of --update-cache  to --no-cache in the Alpine Docker file broke the build.
I can't build an Alpine Docker image on Raspberry Pi 5 using Ubuntu unless I revert that change.
@grasdk Would you kindly opine if this was an oversight?

grasdk:
> To me it seems that the change of --update-cache to --no-cache in the Alpine Docker file broke the build. I can't build an Alpine Docker image on Raspberry Pi 5 using Ubuntu unless I revert that change. @grasdk Would you kindly opine if this was an oversight?

@jdavidpeter That seems weird, but if that is your experience, I can't contest it. `buildx` builds it for ARM64 which I believe is the rPi5 architecture. Can you install and run the edge image from docker hub?
docker pull bpatrik/pigallery2:edge-debian-bookworm

bpatrik:
As far as I can tell buildx in github action works: https://hub.docker.com/r/bpatrik/pigallery2/tags/ last build was 2 months ago on all platforms

PR #954, opened 2024-11-12
History fix

roastedGhosty:
Resolves https://github.com/bpatrik/pigallery2/issues/819.
Sets a parameter on gallery navigation to replace the history state on every navigation change. Thus when navigating phots in the lightbox each photo view acts as the most recent navigation state and when hitting back, it returns to the album.

### DIFF
--- a/src/frontend/app/ui/gallery/grid/grid.gallery.component.ts
+++ b/src/frontend/app/ui/gallery/grid/grid.gallery.component.ts
@@ -193,7 +193,6 @@ export class GalleryGridComponent
   photoClicked(media: MediaDTO): void {
     this.router.navigate([], {
       queryParams: this.queryService.getParams({media}),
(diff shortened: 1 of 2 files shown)


bpatrik:
That is an intended behavior so you can copy the url and it will show the exact same photo that you have open.

roastedGhosty:
Hi can you clarify what you are responding to? Nothing about this code change undoes the behavior you are referring to. As posted in the issue thread here is a gif of how this code change looks (the gif takes a few seconds to load):
firefox_1jlK57aij7
The URL changes with each image open and when hitting browser back we see that the lightbox is closed.

bpatrik:
Ahh okay thank you!

I have some memories that I tried to crack this problem before but failed, but I cant remember why. Based on the gif it does look good, so I will merge it in.

PR #960, opened 2024-12-18
Enable TLS 1.3 in nginx.conf

DavidSchinazi:
TLS 1.3 has been supported in nginx since 2018. This change brings nginx.conf to parity with what nginx recommends by default.


bpatrik:
Thank you!

PR #975, opened 2025-03-11
Provide more reasonable quality for webp video thumbnails

joef42:
qscale=4 provides bad quality for webp video thumbnails. qscale=50 creates better results and has a comparable size to image thumbnails.


bpatrik:
thanks!

Issue #941, opened 2024-10-16
cannot download videos


Berschi:
When trying to download a video, it doesn't download the video but some kind of screenshot of the video. This can be reproduced on the Demo server (I also tested it on the latest docker edge image):
https://pigallery2.onrender.com/gallery?p=bunny.mp4

![image](https://github.com/user-attachments/assets/b0923d39-af14-4558-a109-83494db40b77)

![image](https://github.com/user-attachments/assets/378ce6f8-f75d-4918-a2d1-8468f32225ac)

![image](https://github.com/user-attachments/assets/edbd0597-cb9a-4c13-b6dc-78b43d175ba9)

This was done with Firefox. With the Vivaldi browser, it downloads a not playable mp4 file, size also just 3kb:

![image](https://github.com/user-attachments/assets/691e294c-09b4-41e6-902c-9501147b891c)

skatsubo:
Thank you @Berschi for your report. I stumbled across this issue today too.
The fix, it seems, is a one-character change. It fixed the issue for me. Though I did not test it extensively, so it might cause regressions.

I've created PR. Not sure if it will be merged soon as there is no activity from Patrik since January (:sad_panda:)

You can try the fix by monkey patching the js file in your PG2 container:
```sh
docker compose exec pigallery2 sed -i 's/size?/size/' /app/src/backend/routes/GalleryRouter.js
docker compose restart pigallery2
```

PR #992, opened 2025-06-05
Fix video download

skatsubo:
Fix https://github.com/bpatrik/pigallery2/issues/941 "cannot download videos".

Avoid clashes between `addGetVideo` and `addGetVideoThumbnail` by fixing API endpoint signature. (Compare to `addGetImage` and `addGetResizedPhoto` which work as expected)
Otherwise, `addGetVideoThumbnail` intercepts requests to `.../gallery/content/video.mp4` and returns a thumbnail pic instead of the original video file.


skatsubo:
Alternative solution might be changing the route definition order similar to https://github.com/bpatrik/pigallery2/pull/827 ?

bpatrik:
Thanks! I think it should also work this way.

PR #995, opened 2025-06-10
fix #989 Custom sort order does not work

### LINKED ISSUE #989: Customs sort order does not work
When I apply [#177 ] it has no effect on the sorting of the folders.

1. I set the default sort order in the settings to Descending by name.
2. I create a empty file named '.order_ascending_name.pg2conf' in a subfolder of my top level folder.
3. Result is that nothing changes, the sort order is still the default Descending by name.

I have emptied the browser cache, logged in and out of piGallery2, cleared browser history and temporary files etc but still no change of sort order for that folder.

 - OS: Linux on Raspberry Pi.
 - Browser: Firefox and Edge on Windows 11. Firefox, Chrome, Edge on Android.

**Used app version**:
- docker-latest app version 2.0.0

kagahd:
See https://github.com/bpatrik/pigallery2/issues/989

### DIFF
--- a/src/common/PG2ConfMap.ts
+++ b/src/common/PG2ConfMap.ts
@@ -14,8 +14,8 @@ Utils.enumToArray(SortByTypes).forEach(kv => {
     PG2ConfMap.sorting['.order_random.pg2conf'] = {method: kv.key, ascending: null} as SortingMethod;
     return;
   }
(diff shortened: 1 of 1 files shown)


bpatrik:
Thank you!

PR #1000, opened 2025-06-13
Fixbuild

kagahd:
The build failed due the deprecated `actions/upload-artifact@v3` which this PR fixes by replacing it by `actions/upload-artifact@v4`.
Moreover, to enable forked repositories to push the docker images to **their own** docker hub namespace, the namespace `bpatrik` is replaced by  `${{ secrets.REGISTRY_NAMESPACE }}`, so the users just need to add a github secret named `REGISTRY_NAMESPACE` with value of their docker hub namespace. It's not a secret per se but doing so aligns with the existing secrets `REGISTRY_USERNAME`and `REGISTRY_PASSWORD`.
I also added a `workflow_dispatch` trigger because it's handy to be able to run the workflow on demand and not just on commit.
Last but not least I removed the `rm /var/cache/apk/*` because first it may break the workflow if the github runner is still using the cache and secondly because it's not even needed since the option `--no-cache` is used.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -66,7 +66,7 @@ jobs:
       with:
         start: npm run start-e2e-server
       # after the test run completes store screenshots
-    - uses: actions/upload-artifact@v3
+    - uses: actions/upload-artifact@v4
       if: failure() # Store screenshots only on failures
       with:
          name: cypress-screenshots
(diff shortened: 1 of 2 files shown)


skatsubo:
Hey @kagahd, great contributions! I saw your recent PRs and [pigallery2-metadata-editor](https://github.com/bpatrik/pigallery2/discussions/994), that's awesome. 
It's so unfortunate that Patrik is busy these days and cannot merge them. Seems there is no activity since January 2025, hopefully all is well.

bpatrik:
thanks for the fix!

Let me see if merging them fixes the build

Issue #1047, opened 2025-10-10
Show rating overlay for images in grid view


Berschi:
I would love to have an overlay on the images in the grid view to see its rating.
Example:
https://pigallery2.onrender.com/gallery/?p=IMG_5910.jpg
This image has a rating of 4
<img width="3646" height="1813" alt="Image" src="https://github.com/user-attachments/assets/d2b45a16-ffd5-4b35-8322-c51e06c29311" />

In the grid view I would love to see something like this:
<img width="3803" height="1034" alt="Image" src="https://github.com/user-attachments/assets/6ce8dadc-5d6c-46a8-8ce6-a87b130dbd3e" />
options could be "show never", "show on hover", "show always"

alternatively the rating could be added to the hover bar of the image in the grid view, with the same options as mentioned above:
<img width="977" height="759" alt="Image" src="https://github.com/user-attachments/assets/d289d316-b2e0-4fe7-8d20-26aec98949c9" />

bpatrik:
this should be able to do with an extension.

Berschi:
you are right. I managed to create a very simple extension that shows stars in the grid view:

<img width="1510" height="864" alt="Image" src="https://github.com/user-attachments/assets/85f1cd71-ea62-49c3-8105-04cacb566190" />

So thank you very much :-)

bpatrik:
wow. looks great. 
If you add your extension to https://github.com/bpatrik/pigallery2/blob/master/extension/REPOSITORY.md, others will be able to download from the app

PR #1067, opened 2025-11-07
fixed added entry for Grid Rating Overview extension in REPOSITORY.md

Berschi:
fixed the markdown formatting in REPOSITORY.md

see #1047


Berschi:
@bpatrik sorry I had no time before, but now I set up the main project and tried to run pigallery2 before my last merge (Commit 462d6fdb1773fdb92bd680da4c5122b90cc8b7ba), but there I get the same error when clicking on "+ Install extensions":
image

bpatrik:
The problem is in the repository file itself. App freshly loads from the repo it's not built into the project app

Berschi:
ah that explains it, so because the file in the current main branch is broken, it won't work anywhere...
so there is actually no way of testing that, the REPOSITORY.md file just has to be correctly formatted and work.
then again, sorry about that, I hope this merge request fixes it

bpatrik:
Let's try. If not I will just fix it in the next few days. It's all good :)

Berschi:
yes, now it works :)

bpatrik:
thank you!

I also added a test, so in the future, we would at least know if it breaks things.

PR #1071, opened 2025-11-10
Trigger garbage collector every 1000 files in DiskManager

evrial:


bpatrik:
Thanks. it looks good.

PR #1077, opened 2025-11-19
Build versions between 2.0.0 and 3.0.0 with arm/v7 support

grasdk:
Added ability to build and publish an image using the `build-from-commit_id-between-v2-and-v3` github actions workflow.

This github actions workflow was created in order to retrospectively create builds and publish releases of commits between versions 2.0.0 and 3.0.0, which was the period between October 2023 and late June 2025. The workflow will not work on commits in July 2025 and onwards, which was where `arm/v7` support was dropped. Any release made with the workflow, should still support `arm/v7`, e.g. Raspberry Pi 2 and 3, which was the primary reason to create the workflow.

#### Preconditions

It requires that you set these github secrets for your fork (info from your docker-hub account):
- `REGISTRY_NAMESPACE`
- `REGISTRY_USERNAME`
- `REGISTRY_PASSWORD`

#### Usage

When you run the workflow / github action, it will ask you for a commit ID which will then be built and published to docker hub, given the 3 values above.

Note. The versioning of the artifact are still semantic versioning compatible. They are versioned as 3.0.0-pre_alpha releases with the time and commit ID as part of the name.

Screenshot of the version in the deployed application under settings:
billede

Examples of releases (in my docker-hub repo):
https://hub.docker.com/repository/docker/grasdk/pigallery2/tags

### DIFF
--- a/docker/CONTRIBUTING.md
+++ b/docker/CONTRIBUTING.md
@@ -33,7 +33,7 @@ $ docker buildx build \
   --output=type=docker,dest=pigallery2-test
 ```
 
-### Building a docker image from a specific commit ID
+### Building a docker image from a specific commit ID between version 2.0.0 and 3.0.0
 
 When you fork this repo, you can build and publish an image using the `build-from-commit_id-between-v2-and-v3` github actions workflow.
 


bpatrik:
Given that its solved for now. I will reject this PR. sorry.

PR #1089, opened 2025-12-01
Use patched version of exifr

evrial:
All exifr tests are passing
```
npm test
783 passing (955ms)
```


bpatrik:
No need for this. Exif closes the file: https://github.com/MikeKovarik/exifr/blob/6cbf6e921688faf7723e1f2e0b9e672d1f0aa21c/src/Exifr.mjs#L37

Issue #1102, opened 2025-12-21
Cant docker build using supplied Dockerfile


minsonj2016:
I was trying to do a docker build using the alpine Dockerfile and cant get past this error

0.616 Error: Cannot find module '/app/src/backend/index'

I tried 3.0.1 , 3.1.0 and latest

Its not a big deal as the 'pulled' image works fine . I was just trying to learn a little more about docker/npm/pigallery2

Thanks

bpatrik:
Which dockerfile are you using? My guess is that you are missing the source file (if i remember correctly the build source should be next to the docker file)

minsonj2016:
/nt01/Data/PiGallery2/pigallery2-master/docker/alpine $ ls
Dockerfile.build

bpatrik:
yes you need the source files in the same folder: https://github.com/bpatrik/pigallery2/blob/3c3fe13271487140ccdad1386f23245f4c6e5671/docker/alpine/Dockerfile.build#L28

minsonj2016:
Let me clarify what 'm doing .
I wget latest and unzip it.
Now I have pigallery2-master in my current directory.
Then I copy the alpine dockerfile to .
I then modify the Dockerfile changing "COPY pigallery2-release /app" to "COPY pigallery2-master /app"
Then docker build -t pigallery:local-latest .

[](/mnt01/Data/PiGallery2 $ docker build -t pigallery:local-latest .
[+] Building 196.5s (14/14) FINISHED                                                                                                                      docker:default
 => [internal] load build definition from Dockerfile                                                                                                                0.0s
 => => transferring dockerfile: 2.69kB                                                                                                                              0.0s
 => [internal] load metadata for docker.io/library/node:22-alpine3.22                                                                                               0.2s
 => [internal] load .dockerignore                                                                                                                                   0.0s
 => => transferring context: 2B                                                                                                                                     0.0s
 => [internal] load build context                                                                                                                                   2.0s
 => => transferring context: 123.71MB                                                                                                                               2.0s
 => [builder 1/6] FROM docker.io/library/node:22-alpine3.22@sha256:b2358485e3e33bc3a33114d2b1bdb18cdbe4df01bd2b257198eb51beb1f026c5                                 0.0s
 => CACHED [builder 2/6] RUN apk add --no-cache --update     --repository=https://dl-cdn.alpinelinux.org/alpine/edge/main     --repository=https://dl-cdn.alpineli  0.0s
[...]

PR #1127, opened 2026-01-29
Updated Dockerfile

moaz-ahmed:
Changed builder path to correctly refer to `/build/release` instead of `/app` which doesn't exist. This has caused the selfcontained docker file build to [fail](https://github.com/bpatrik/pigallery2/issues/1102).


bpatrik:
thanks!

PR #1131, opened 2026-02-08
Fix biased random sort by using Fisher-Yates shuffle

infinitejones:
## Summary

- The random sort option used `Array.sort()` with a random comparator, which is a [well-known antipattern](https://blog.codinghorror.com/the-danger-of-naivete/) that produces biased results — items tend to stay near their original (alphabetically sorted) position
- This is especially noticeable when filenames share a common prefix (e.g. `France - beach`, `France - city`, `Italy - beach`, `Italy - city`), as the alphabetical pre-sort clusters them and the biased shuffle preserves those clusters
- Replaced with [Fisher-Yates shuffle](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle) for both media and directories, which guarantees a uniform random permutation
- The existing seeded PRNG is still used, so the shuffle remains deterministic within a page load

## Test plan

- [ ] Select random sorting on a gallery with many similarly-named files and verify items are well-mixed
- [ ] Verify the shuffle is stable within a page load (navigating away and back gives the same order)
- [ ] Verify directory random sorting also works correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/frontend/app/ui/gallery/navigator/sorting.service.ts
+++ b/src/frontend/app/ui/gallery/navigator/sorting.service.ts
@@ -173,18 +173,10 @@ export class GallerySortingService {
         break;
       case SortByTypes.Random:
         this.rndService.setSeed(media.length);
-        media.sort((a: PhotoDTO, b: PhotoDTO): number => {
-          if (a.name.toLowerCase() < b.name.toLowerCase()) {
-            return -1;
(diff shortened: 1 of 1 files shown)


bpatrik:
thanks!

PR #1139, opened 2026-03-24
Add Apple Live Photo Support

### LINKED ISSUE #609: Support for Apple Live Photos

bexelbie:
Closes #609

## What

Apple Live Photos (HEIC still + MOV video pairs) are displayed as a single media item with inline video playback, instead of appearing as two separate
entries.

The feature is gated behind `Client.Media.LivePhoto.enabled` (default: true).  Schema version 41→42 triggers a full re-index on upgrade.

This is my first commit.  I believe I have followed the guide in the code and met your standards.  I'm happy to make changes.

## Strategy

- **Index everything, pair at response time.** Both the HEIC and MOV are indexed independently. Pairing happens in `cleanUpGalleryResults` when building the API response — no schema-level joins or foreign keys.
- **Metadata-based matching.** Pairs are linked by a Content Identifier UUID embedded in both files (Apple MakerNote tag 0x0011 in HEIC, QuickTime `content.identifier` in MOV). No filename heuristics.
- **Companion videos are filtered from listings** after pairing, so the grid shows one item per Live Photo.
- **`contentIdentifier` stays server-side only.** It lives on the DB entity for indexing but is stripped before the API response — the client never sees it.

## UX

- **Grid:** A "LIVE" badge appears on paired photos.
- **Lightbox:** Hovering (or touching) the LIVE badge plays the companion video inline, muted, matching Apple's behavior. The badge is the only interactive hotspot — it does not interfere with existing swipe/zoom/pan gestures.
- **Info panel:** Shows both the still image and companion video metadata (filename, dimensions, size, duration) in separate rows.

## Commits

1. **Data model, config, metadata extraction** — `contentIdentifier` column, `LiveVideoInfo` interface, MakerNote parser, ffprobe extraction, config toggle, schema version bump (41→42)
2. **Response-time pairing middleware** — `pairLivePhotos()` in GalleryMWs, `contentIdentifier` stripping, ContentWrapper pack/unpack for `liveVideoPath`/`liveVideoInfo`
3. **Lightbox playback** — hover-to-play on badge, `pointer-events: none` container to avoid gesture conflicts, muted video element
4. **Grid badge + info panel** — LIVE indicator in grid, companion video details in info panel with videocam icon

## Tests

- **Backend:** MetadataLoader extraction tests (MOV content ID, HEIC MakerNote, non-Live-Photo passthrough) and GalleryMWs pairing tests (matching, non-matching, config disabled, search results, unpaired video retention, contentIdentifier stripping)
- **Frontend:** GridMedia unit tests, grid photo badge rendering, lightbox media rendering + interaction (mouse and touch events), info panel companion video display


bexelbie:
Untitled


Untitled 2

test/backend/assets/live_photo/IMG_7943.HEIC
bpatrik, line 1: if possible make these assets smaller

src/frontend/app/ui/gallery/lightbox/media/media.lightbox.gallery.component.html
```diff
@@ -15,6 +15,30 @@
...
+           playsinline
+           (error)="onImageError()"
+           #liveVideo>
+      <source [src]="liveVideoSrc" (error)="onSourceError()">
+    </video>
+    <div class="live-photo-badge"
+         [class.active]="liveVideoPlaying"
+         (mouseenter)="startLiveVideo()"
+         (mouseleave)="stopLiveVideo()"
+         (touchstart)="startLiveVideo()"
```
bpatrik: does this mean that if i click the button it will start playing the video?

if not i think that would be nice.

src/common/entities/MediaDTO.ts
```diff
@@ -4,12 +4,23 @@ import {FileDTO} from './FileDTO';
 import {SupportedFormats} from '../SupportedFormats';
 import {Utils} from '../Utils';
 
+export interface LiveVideoInfo {
```
bpatrik: If possible, can you make this and video metadata interface interface have the same parent/ share properties?

src/common/config/public/ClientConfig.ts
```diff
@@ -1381,6 +1381,20 @@ export class ClientGPXCompressingConfig {
   enabled: boolean = true;
 }
 
+@SubConfigClass({tags: {client: true}, softReadonly: true})
+export class ClientLivePhotoConfig {
+  @ConfigProperty({
+    tags: {
+      name: $localize`Enable`,
+      priority: ConfigPriority.advanced,
+      uiResetNeeded: {db: true},
```
bpatrik: To my understanding this feature reads the identifier and saves to the DB.
At serving time it pairs photos with videos. 

It think we should always save the ID to the DB and only gate this feature during serving time, so its less confusing to the user when they switch this feature on and off. 

make sure that the ID is filtered in the response.

src/backend/model/fileaccess/MetadataLoader.ts
```diff
@@ -179,7 +188,7 @@ export class MetadataLoader {
       creationDate: 0,
       fileSize: 0,
     };
-    const exifrOptions = {
+    const exifrOptions: any = {
```
bpatrik: try to avoid using any, lint will fail on it. maybe set makerNote to false instead as default? Ot just to Config.Media.LivePhoto.enabled

bpatrik (commented):
Thank you for the contribution!

I'm generally happy with it. I made some comments, can you please take a look?


Also are mov files support on non mac ?
Like linux/win + FF or chrome users will be able to make this feature work?

I only found an old thread that its not straightforward: https://stackoverflow.com/questions/57060193/browser-support-for-mov-video

src/frontend/app/ui/gallery/grid/GridMedia.ts
```diff
@@ -29,6 +31,27 @@ export class GridMedia extends Media {
...
+    const encodedPath = encodeURI(this.media.liveVideoPath)
+      .replace(new RegExp('#', 'g'), '%23')
+      .replace(new RegExp('\\$', 'g'), '%24')
+      .replace(new RegExp('\\?', 'g'), '%3F');
+    return Utils.concatUrls(
+      Config.Server.urlBase,
+      Config.Server.apiPath,
+      '/gallery/content/',
+      encodedPath,
+      '/bestFit'
```
bpatrik, line 51: I guess this will let ppl to transcode the mov files and make them work on all browsers

bexelbie:
I believe this code (I squashed the commits) addresses all of your feedback, including smaller test assets, the click to play changes, shared interface, and always extracting as well as lint `any types` clean up.  Regarding mov files, as there is no on-the-fly video conversion, users on Linux will need to have done a transcoding job.  If they have, they will get served the transcoded file, just like with other videos.  Otherwise, they will get the feedback they would normally get.

WDYT?

bpatrik:
Sounds good lets give it a try :)

thank you for the contribution

PR #1151, opened 2026-05-20
Read metadata from sharp fallback for image formats

Railline:
## Summary
- read embedded EXIF/XMP/IPTC buffers exposed by sharp when exifr cannot parse the file directly
- keep parsed sharp metadata sections together before mapping them into PiGallery metadata
- add a focused WebP regression fixture local to the metadata loader test
- update affected tests for newly indexed GIF media and nondeterministic SQL cover selection

## Tests
- npm run build-backend
- npm run lint
- mocha test/backend/unit/model/threading/MetaDataLoader.spec.js
- mocha test/backend/unit/model/threading/DiskManagerWorker.spec.js
- mocha test/backend/unit/model/sql/IndexingManager.spec.js --grep should save photos with extreme parameters with local MariaDB


bpatrik:
Sorry but what does the PR solve exactly? The app was already falling back to sharp

Railline:
> Sorry but what does the PR solve exactly? The app was already falling back to sharp

Hi hi ^^

This PR fixes WebP metadata extraction in the Sharp fallback path.

The app already falls back to Sharp when exifr cannot parse the image directly, but Sharp can expose raw exif, xmp, and iptc buffers. Previously those buffers were not parsed separately, so metadata such as creation date, timezone offset, keywords, and face regions could be missed.

The PR parses those buffers and maps the extracted data into PiGallery2 metadata.

bpatrik:
sounds good thanks for the contribution

test/backend/unit/model/sql/IndexingManager.spec.ts
```diff
@@ -495,6 +495,8 @@ describe('IndexingManager', (sqlHelper: DBTestHelper) => {
 
     DirectoryDTOUtils.removeReferences(selected);
     removeIds(selected);
+    delete selected.cache.cover;
```
bpatrik, line 498: Im not sure why do you need to remove this. I would expect that you change does not touch this part

test/backend/unit/model/threading/MetaDataLoader.spec.ts
```diff
@@ -312,7 +318,7 @@ describe('MetadataLoader', () => {
     for (let i = 0; i <= 8; ++i) {
       it('Landscape ' + i, async () => {
         const data = await MetadataLoader.loadPhotoMetadata(path.join(__dirname, '/../../../assets/orientation/Landscape_' + i + '.jpg'));
-        const expected = require(path.join(__dirname, '/../../../assets/orientation/Landscape.json'));
+        const expected = Utils.clone(require(path.join(__dirname, '/../../../assets/orientation/Landscape.json')));
```
bpatrik, line 321: Im not sure why do you need to remove this. I would expect that you change does not touch this part

Railline:
my bad i have test my build and that make more problem than it solve

PR #1156, opened 2026-05-30
Revert "Add Apple Live Photo Support"

bpatrik:
Reverts bpatrik/pigallery2#1139 as it is breaking tests


bpatrik:
not this one was the culprit

PR #1161, opened 2026-06-05
fix: display HEIC/DNG/ARW/TIFF photos in non-Safari browsers

vqminh:
Two issues fixed:

1. Convert non-browser-native formats to WebP for full-res viewing. HEIC, DNG, ARW, and TIFF files were served raw to the browser, but only Safari supports HEIC natively. Now these formats are converted to full-resolution WebP on first access and cached in TranscodedFolder.

2. Check file size when validating cached thumbnails. Previously the code used fsp.access()/fs.existsSync() which only checks existence. Failed conversions that produced 0-byte files were treated as valid thumbnails, permanently breaking those photos. Now all cache checks verify stat.size > 0.

### DIFF
--- a/src/backend/middlewares/thumbnail/ThumbnailGeneratorMWs.ts
+++ b/src/backend/middlewares/thumbnail/ThumbnailGeneratorMWs.ts
@@ -93,7 +93,14 @@ export class ThumbnailGeneratorMWs {
           size
         );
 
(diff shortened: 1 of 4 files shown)


bpatrik:
I think we do this already. or at least partially. There might be some issue when zooming in

PR #1162, opened 2026-06-07
feat: make photo title searchable (any_text + title: prefix)

cat101:
As with my other PRs, this has been generated using AI. This one is really a small missing feature.  I have manually tested it on my environment.

------------------------
## Problem

Photo titles (`Xmp.dc.title` / `IPTC ObjectName` / `Photoshop Headline` / `acdsee.caption`) are extracted by `MetadataLoader.mapTitle` and stored on `media.metadata.title` — but `SearchManager` never consults that column. Titles set in digiKam et al. are loaded into the DB and shown in the info panel (added in #798), yet are invisible to search: a bare word in the search box doesn't find them, and there's no `title:` prefix.

Reproducer: tag a photo with title "Shuttle" in digiKam → re-index → search "Shuttle" → no results.

## Fix

Add a `SearchQueryTypes.title` text-search type alongside the existing caption / file_name / etc., wired into:

- **SearchManager** — new autocomplete branch + new where-builder clause on `media.metadata.title`, mirroring the existing `caption` clause. The negated branch wraps the predicate in a `Brackets` that also matches `IS NULL`, since title is sparse and `NULL NOT LIKE 'x'` is NULL under three-valued logic (would otherwise drop untitled photos from negated `any_text` results).
- **SearchQueryParser** — `title` in `QueryKeywords` + `defaultQueryKeywords` (`'title'`) + stringify switch. Parsing auto-picks up the prefix because it iterates `TextSearchQueryTypes`.
- **AutoCompleteItemsPerCategoryConfig.title** (default `3`) — matches caption.
- **Frontend** — `EnumTranslations` label + autocomplete icon (reuses `ionTextOutline`).

`any_text` now also matches `media.metadata.title`, fixing the original bug. The DB schema already has the `title` column (added in #798), so no migration is required.

## Tests

- `SearchQueryParser` round-trip for `title` type.
- `SearchManager` — new `it('as title')`: asserts both `title:Shuttle` (explicit) and `any_text:Shuttle` (the bug fix) return the seeded photo.
- `TestHelper.getPhotoEntry2` (sw2.jpg) gets `metadata.title = 'Shuttle'`. Picked because "Shuttle" contains no substring exercised by existing autocomplete assertions (`tat`, `star`, `wars`, `phantom`, `arch`, `wa`, `a`, `sw`, `han`) so no other test had to be updated.

## Notes

- Diff: 9 files, +101/−1. No new dependencies. No schema migration.
- Backwards-compatible: existing search behavior unchanged when title is empty/NULL.
- The `IS NULL` handling in the negate branch is a stricter pattern than the existing `caption` clause uses — caption has the same latent NULL bug, but I left it alone to keep the diff minimal and focused. Happy to file a follow-up if you want it normalized.
- Frontend i18n: only the English label was added (`$localize\`Title\``). The .xlf translation files would need a follow-up pass.

### DIFF
--- a/src/backend/model/database/SearchManager.ts
+++ b/src/backend/model/database/SearchManager.ts
@@ -313,6 +313,37 @@ export class SearchManager {
 
     if (
       type === SearchQueryTypes.any_text ||
+      type === SearchQueryTypes.title
+    ) {
+      const q = photoRepository
+        .createQueryBuilder('media')
+        .select('DISTINCT(media.metadata.title) as title')
+        .where(
+          'media.metadata.title LIKE :value COLLATE ' + SQL_COLLATE,
+          {value: '%' + value + '%'}
+        );
+
+      if (session.projectionQuery) {
+        if (session.hasDirectoryProjection) {
+          q.leftJoin('media.directory', 'directory');
+        }
+        q.andWhere(session.projectionQuery);
+      }
+      q.limit(
+        Config.Search.AutoComplete.ItemsPerCategory.title
+      );
+      partialResult.push(
+        this.encapsulateAutoComplete(
+          (
+            await q.getRawMany()
+          ).map((r) => r.title),
+          SearchQueryTypes.title
+        )
+      );
+    }
+
+    if (
+      type === SearchQueryTypes.any_text ||
       type === SearchQueryTypes.directory
     ) {
       const dirs = await directoryRepository
@@ -1021,6 +1052,16 @@ export class SearchManager {
       ) {
         q[whereFN](
           getLikeExpr('media.metadata.caption', 'text'),
+          textParam
+        );
+      }
+
+      if (
+        (query.type === SearchQueryTypes.any_text && !directoryOnly) ||
+        query.type === SearchQueryTypes.title
+      ) {
+        q[whereFN](
+          `media.metadata.title ${LIKE} :text${queryId} COLLATE ${SQL_COLLATE}`,
           textParam
         );
       }
--- a/src/backend/model/database/SearchManager.ts
+++ b/src/backend/model/database/SearchManager.ts
@@ -1060,9 +1060,19 @@ export class SearchManager {
         (query.type === SearchQueryTypes.any_text && !directoryOnly) ||
         query.type === SearchQueryTypes.title
       ) {
+        // title is sparse: most photos have no title set. When negating, treat
+        // NULL as a match so a `-title:X` (or negated any_text) doesn't filter
+        // out untitled photos due to NULL NOT LIKE = NULL three-valued logic.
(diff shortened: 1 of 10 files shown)


bpatrik:
Thanks!

please add the new filed to the content wrapper also with proper alias. 

Also please merge as I did some significant changes to the search manager recently

cat101:
Both comments adressed

------- AI comment below -------

Thanks for the review — both points are done, plus one thing I found on the way.

**Content wrapper.** `title` was absent from `ContentWrapper` entirely, so it was travelling under its full key while every sibling metadata field packs to a single letter. Added as `'h'` with the matching unpack (c7f7c1e).

One deliberate asymmetry there, flagged because it looks like an oversight otherwise: the video branch does `delete metadata.caption`, and I did **not** do the same for `title`. `loadVideoMetadata()` runs `mapMetadata()` over an XMP sidecar, which sets both, so a video can genuinely carry a title — dropping it would make that title findable by `title:` search while the UI showed nothing. Happy to change it if you'd rather they match.

**Rebased** onto current master (cd42146).

**And the NULL fix was incomplete — that's the commit worth your attention** (d09356e). My earlier `title IS NULL` guard was right in reasoning and applied to one of the three places that need it:

- `metadata.caption` is nullable and sits four lines above `title` in the same branch, with no guard.
- `positionData.country` / `state` / `city` are all `nullable: true` and had none either.

Since `any_text` ORs these together, negating it ANDs the negated parts, so one unguarded nullable column empties the whole result. Concretely: `NOT any_text:(X)` returns **nothing** in any library where some photo has no caption — which is every library — and `NOT position:(X)` silently drops every photo that has no location. So the fix I claimed in the previous commit message wasn't actually delivered until this one.

The three open-coded variants are now one `matchNullableField()` helper, so the next nullable text column can't be added without the guard.

Your own test suite already had the fixture that proves it: `getVideoEntry()` leaves `caption` null and sets no `positionData` at all, so `swVideo.mp4` was missing from `NOT any_text:('Boba')` and the test asserted its absence. It's in the expected set now, and the adjacent "all should have faces" loop skips it — a video carries no face regions.

I've kept the caption/position guard as a **separate commit** from the content-wrapper change precisely so you can cherry-pick or ask me to split it into its own PR — it's a pre-existing bug rather than anything to do with title search, and I'd understand wanting it reviewed on its own.

**One unrelated observation while verifying.** `test/backend/unit/model/sql/SearchManager.spec.ts` has 9 failing tests on pristine `master` (cd42146) under sqlite — 34 passing / 9 failing on a bare checkout, nothing to do with this branch. With this PR it's 35 / 9. Mentioning it in case it's news; happy to open a separate issue with the list if useful.

PR #1168, opened 2026-06-22
Support glob wildcards in excludeFolderList

### LINKED ISSUE #192: Exclude folder based on regexp when indexing
Pigallery2 indexes all folders recursively starting from root, which might not be ideal for mixed content.

Excluding a user given list of file or folder during indexing. It would be great if the list can be given as regular expressions.

cycsmail:
Fixes #192.

`excludeFolderList` only matched folder names exactly, so there was no way to exclude e.g. all `temp*` folders at once. This makes the no-slash case match with glob patterns (`*` and `?`), reusing the same `globToRegex` helper the filename exclude list already uses. Names without wildcards still match exactly, so the existing defaults (`.dtrash`, `$RECYCLE.BIN`, ...) keep working as before.

Added unit tests for the glob matching and the plain-name case, and updated the setting description to mention the wildcards. The DiskManagerWorker spec passes.

### DIFF
--- a/src/backend/model/fileaccess/DiskManager.ts
+++ b/src/backend/model/fileaccess/DiskManager.ts
@@ -104,7 +104,7 @@ export class DiskManager {
           return true;
         }
       } else {
-        if (exclude === dir.name) {
+        if (DiskManager.globToRegex(exclude).test(dir.name)) {
           return true;
         }
       }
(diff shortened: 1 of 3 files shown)


cycsmail:
this one's still open if you get a chance, happy to rebase onto master or trim it down if you'd rather. the 22.x job is the same 6h hang as on #1175, not related to the change.

PR #1175, opened 2026-07-15
Hide the mouse cursor when the lightbox controls are hidden

### LINKED ISSUE #736: Mouse cursor in fullscreen mode
pigallery2 automatically hides the control buttons when detects that I have switched to auto play after a few seconds of inactivity. This helps avoid distraction and focus on the task at hand. 

But, the mouse cursor will continue to show, even in fullscreen. 

capture_20231007_072802

cycsmail:
Fixes #736

The cursor stayed visible in the lightbox even after the controls faded out (e.g. during a slideshow), so it sat there on top of the photo. This binds a `hide-cursor` class to `#swipeable-container` off the existing `controllersDimmed` state and adds `cursor: none` for it, so the cursor fades away together with the controls and comes back as soon as they reappear.

Added a small component spec covering both states (cursor hidden while dimmed, visible while shown); it passes.

### DIFF
--- a/src/frontend/app/ui/gallery/lightbox/controls/controls.lightbox.gallery.component.css
+++ b/src/frontend/app/ui/gallery/lightbox/controls/controls.lightbox.gallery.component.css
@@ -302,6 +302,11 @@ input[type="range"].zoom-progress::-moz-range-track {
   opacity: 0.0;
 }
 
+/* hide the mouse cursor together with the controls (e.g. during slideshow) */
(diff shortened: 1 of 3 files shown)


cycsmail:
looks like the 22.x test job hung and got killed at the 6h limit, doesn't seem related to the change (it's a css/class binding). mind re-running it?

bpatrik:
Thanks!

PR #1177, opened 2026-08-02
Make boolean search keywords case-insensitive

### LINKED ISSUE #1090: bug: Search query language is case sensitive for boolean keywords
searching for `a or b` is not the same as `a OR b`

cycsmail:
Fixes #1090

The query parser only matched the `and`/`or` boolean keywords with exact casing, so `OR` and `AND` (or mixed case like `Or`) weren't recognised. I added a small case-insensitive variant of `startsWith` and use it when matching those keywords during tokenization.

Added a unit test covering the case-insensitive variants. The SearchQueryParser unit tests pass locally (19 passing).

### DIFF
--- a/src/common/SearchQueryParser.ts
+++ b/src/common/SearchQueryParser.ts
@@ -218,8 +218,13 @@ export class SearchQueryParser {
     // tokenize
     const tokenEnd = fistSpace();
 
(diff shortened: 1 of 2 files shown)


bpatrik:
Thank you!

Issue #557, opened 2022-11-20
Allow user to create album


sam365724:
Can only admins create albums based on searches? I'd like an ordinary user to be able to create albums, is that possible?

bpatrik:
At the moment Admin role is hardcoded:
https://github.com/bpatrik/pigallery2/blob/9b56331f509da9fb1f5a8225030a845d31f13cda/src/frontend/app/ui/albums/albums.component.ts#L44-L46

I assume it is also checked on the server side (its been a while wehn I added it), but should not be to hard to enable it.

dasmaetthes:
See PR #1181

PR #1181, opened 2026-08-11
Fix Allow user to create album #557

dasmaetthes:


bpatrik:
Please do not open it up for all users. 

You can add a new minRole for album creation to the config and use that to gate album creation. We do that at multiple places.

dasmaetthes:
I made some changes and added a new minRole. Not sure if it's the correct way.

PR #1182, opened 2026-08-11
Support manual-install extension listings

v-marinkov:
## Summary

- allow extension repository entries without a ZIP URL
- show a **Manual installation** link for those entries instead of a broken Install action
- add parser coverage for a manual-install entry

## Why

Some extensions need host-side setup that PiGallery2's ZIP installer cannot perform. For example, [PiGallery2 Curation Requests](https://github.com/v-marinkov/pigallery2-curation-requests) includes a persistent SQLite workflow, a frontend loader, Docker mounts, and trusted-host review/deletion commands. Advertising a normal branch ZIP would produce an incomplete installation.

This PR adds the generic UI capability first. The repository file is fetched live from master, including by older PiGallery2 releases whose UI still assumes every entry has a ZIP. Adding an empty-Download listing in the same PR would therefore expose a broken Install button to those existing versions.

After this support ships in a PiGallery2 release, I will open a small follow-up PR adding Curation Requests to extension/REPOSITORY.md with an empty Download cell. Supported clients will then direct users to its documented .env-driven install.sh workflow.

## Testing

    npm run test-backend -- --grep ExtensionRepository

Result: 2 passing.

### DIFF
--- a/src/common/entities/extension/ExtensionListItem.ts
+++ b/src/common/entities/extension/ExtensionListItem.ts
@@ -3,6 +3,6 @@ export interface ExtensionListItem {
   name: string;
   url?: string;
   readme?: string;
-  zipUrl: string;
+  zipUrl?: string;
   installed?: boolean;
 }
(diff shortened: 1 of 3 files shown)


bpatrik:
Sounds good to me. thanks!

Also nice extension!

